PR #1, opened 2020-05-15
Initial client

michalpristas:
Matches the one from beats repo

running mage update will generate client code and injects a license header

cc @ph @blakerouse


ph:
When we start creating the new contracts, let's add a changelog so we can communicate ou changes with endpoint.

ruflin:
As I expect all of this code to be under ES license I don't think we need the x-pack prefix, it would only complicate the code structure. And yes, we need the headers.

michalpristas:
updated with proto file license header generation
do we need some common license file here?

ph:
lets get that merged @michalpristas to iterate.

PR #2, opened 2020-05-18
Update the client protocol to new defined contract

blakerouse:
Updates the client protocol to reverse the connection adding the Checkin and Actions bi-directional streams.


elastic-agent-client.proto
```diff
@@ -2,25 +2,105 @@
...
 
-message ConfigResponse {
+// A state expected message is streamed from the Elastic Agent to the application informing the application
+// what Elastic Agent expects the applications state to be.
+message StateExpected {
+    enum State {
+        // Expects that the application is running.
+        RUNNING = 0;
+        // Expects that the application is stopping.
+        STOPPING = 1;
```
michalpristas, line 52: are we expecting it to be stopping or stopped, just a small detail

elastic-agent-client.proto
```diff
@@ -2,25 +2,105 @@
...
+    }
+    // Expected state of the application.
+    State state = 1;
+    // Index of the either current configuration or new configuration provided.
+    uint64 config_state_idx = 2;
+    // Resulting configuration. (If the application already has the current `config_state_idx` this
+    // with be empty.)
```
michalpristas, line 59: `s/with/will` probably

elastic-agent-client.proto
```diff
@@ -2,25 +2,105 @@
...
+    string config = 3;
 }
 
-message StatusRequest{
+// A action request is streamed from the Elastic Agent to the application so an action can be performed
```
michalpristas, line 63: very minor `s/A action/An action`

elastic-agent-client.proto
```diff
@@ -2,25 +2,105 @@
...
+// by the connected application.
+message ActionRequest {
+    // Token that is used to unique identify the application to agent. When agent started this
+    // application it would have provided it this token.
+    string token = 1;
```
michalpristas, line 68: do we need unique token when sending req from agent to application? it should be probably moved to `ActionResponse`

elastic-agent-client.proto
blakerouse: Once stopped the application should be off and the GRPC should be disconnected at that point. So I was going with `stopping` which also goes with `running` both being `-ing` based.

elastic-agent-client.proto
blakerouse: You are correct it would be more appropriate on `ActionResponse` but being that Agent shouldn't even send a ActionRequest until the application has checked in with the correct unique token, then really this can be removed all together (I removed it).

PR #4, opened 2020-05-19
Add a client implementation to communicate with Elastic Agent.

blakerouse:
Adds the `pkg/client` module to encapsulate the communication with Elastic Agent greatly simplifying the need for the application to worry about the protocol implementation with Elastic Agent.

An example implementation using the client would be:

```
type EchoAction struct{}
func (a *EchoAction) Name() string { return "echo" }
func (a *EchoAction) Execute(params map[string]interface{}) (map[string]interface{}, error) {
    return map[string]interface{]{
        "result": params["echo"],
    }
}

type StateHandler struct{}

func (s *StateHandler) OnConfig(config string) {
    fmt.Printf("Update to latest config: %s\n", config)
}

func (s *StateHandler) OnStop() {
    fmt.Printf("Stop\n")
}

func (s *StateHandler) OnError(err error) {
    fmt.Printf("Log error: %s\n, err)
}

func main() {
    s := &StateHandler{}
    c := client.New(":8888", "unique_token", s, []Actions{&MyAction{}}, grpc.WithInsecure())
    err := c.Start(context.Background())
    if err != nil {
        panic(err)
    }
    c.Status(proto.StateObserved_HEALTHY, "Running")
    c.Stop()
}
```

The client has 90.7% test coverage, its only missing test coverage for transient error cases. Testing was also completed with data racing, to ensure that no data race issues existed.


blakerouse:
Forgot to mention that I did have to change the protocol for Actions to add a Token to the `ActionResponse` message. This is because grpc can use connection pooling resulting the same client using multiple connections to the same agent so we need to identify the application to the agent specifically with the Actions streams, just like it is done for the Checkin. I was hoping that we could wait for Checkin then ensure that the same connection is shared between the 2, but that didn't seem to be a really good way of doing that.

So on start the Client will send an initial ActionResponse to the Elastic Agent identifying itself with action ID set to `init`.

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+	"time"
+
+	"google.golang.org/grpc"
+
+	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
+	"github.com/elastic/elastic-agent-client/v7/pkg/utils"
+)
+
+// CheckinMinimumTimeout is the amount of time the client must send a new checkin even if the status has not changed.
+const CheckinMinimumTimeout = time.Second * 30
```
ruflin, line 21: I wonder if we should use a lower value here to be able to report faster if a process is not active anymore?

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+}
+
+// startActions starts the go rountines to send and receive actions
+func (c *Client) startActions() {
+	c.wg.Add(1)
+
+	// results our held outside of the retry loop, because on re-connect
+	// we still want to send the responses that either failed or haven't been
+	// sent back to the agent.
+	actionResults := make(chan *proto.ActionResponse, 100)
```
michalpristas, line 285: why 100? maybe this should be sync and enable processing action after one is finished. sender then should handle timeouts in processing or send message after last one is confirmed, otherwise i think messages might be lost

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+							Token:  c.token,
+							Id:     action.Id,
+							Status: proto.ActionResponse_FAILED,
+							Result: ActionErrUnmarshableParams,
+						}
+						continue
+					}
+
+					// perform the action
+					go func() {
```
michalpristas, line 352: as ^^ should we make this sychronous? what if actions of conflicting nature are performed at the same time?

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+
+			// action responses
+			go func() {
+				defer wg.Done()
+				for {
+					select {
+					case <-done:
+						return
+					case res := <-actionResults:
+						err := actionsClient.Send(res)
```
michalpristas, line 395: as with checking if this fails should we return goroutines and let loop to recreate stream?

pkg/client/client.go
ph: Note: I think this value should be configurable at runtime, for lets say that you want more real time response from the client.

ph: I just reread the above, I don't think my comment make sense in that context.

pkg/utils/utils.go
```diff
@@ -0,0 +1,16 @@
...
+
+package utils
+
+import "encoding/json"
+
+// JSONMustMarshal marshals the input to JSON []byte and panics if it fails.
+func JSONMustMarshal(input interface{}) []byte {
+	res, err := json.Marshal(input)
+	if err != nil {
+		panic(err)
```
ruflin, line 13: panic looks pretty extreme here at the same time we should be in full control for all the messages that are converted ...

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+			actionMap[act.Name()] = act
+		}
+	}
+	return &Client{
+		target:          target,
+		opts:            opts,
+		token:           token,
+		impl:            impl,
+		actions:         actionMap,
+		cfgIdx:          0, // 0 is initial, no config state
```
ruflin, line 111: Perhaps worth a `const`?

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+		observed:        proto.StateObserved_STARTING,
+		observedMessage: "Starting",
+		minCheckTimeout: CheckinMinimumTimeout,
+	}
+}
+
+// Start starts the connection to Elastic Agent.
+func (c *Client) Start(ctx context.Context) error {
+	ctx, cancel := context.WithCancel(ctx)
+	c.ctx = ctx
```
ruflin, line 122: Nit: Why not assign it directly on line 121? Same for cancel?

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+			}
+
+			checkinClient, err := c.client.Checkin(c.ctx)
+			if err != nil {
+				c.impl.OnError(err)
+				continue
+			}
+
+			var wg sync.WaitGroup
+			wg.Add(2)
```
ruflin, line 173: When I was reading this my initial question was: Why `2`. I then found out further down the second wg.Done. I wonder if we should have `wg.Add(1)` on line 176 and the second one line 211. I think it would make the code more readable to have Add and Done closer together.

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+
+// Status updates the current status of the client in the Elastic Agent.
+func (c *Client) Status(status proto.StateObserved_Status, message string) {
+	c.obsLock.Lock()
+	c.observed = status
+	c.observedMessage = message
+	c.obsLock.Unlock()
+}
+
+// startCheckin starts the go rountines to send and receive check-ins
```
ruflin, line 152: Could you add some more details here that this starts 3 go routines and what they do?

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+			wg.Wait()
+		}
+	}()
+}
+
+// startActions starts the go rountines to send and receive actions
+func (c *Client) startActions() {
+	c.wg.Add(1)
+
+	// results our held outside of the retry loop, because on re-connect
```
ruflin, line 282:
```suggestion
	// results are held outside of the retry loop, because on re-connect
```

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+			if err != nil {
+				c.impl.OnError(err)
+				continue
+			}
+
+			// initial connection of stream must send the token so
+			// the Elastic Agent knows this clients token.
+			actionResults <- &proto.ActionResponse{
+				Token:  c.token,
+				Id:     "init",
```
ruflin, line 306: worth a `const`?

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+				Status: proto.ActionResponse_SUCCESS,
+				Result: []byte("{}"),
+			}
+
+			var wg sync.WaitGroup
+			wg.Add(2)
```
ruflin, line 312: Reminds me of my comment further above :-)

pkg/client/client.go
blakerouse: This is only used when the current status of the application has not changed so if the application sets the status to `HEALTHY` that will be sent immediately at that point (minus some debounce of 500ms). Then it will keep sending `HEALTHY` every 30 seconds.

Being that Elastic Agent will have an expectation of check-in time, I don't think we need to have this configurable.

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+						if err != io.EOF {
+							c.impl.OnError(err)
+						}
+						close(done)
+						return
+					}
+
+					if c.expected == proto.StateExpected_STOPPING {
+						// in stopping state, do nothing with any other expected states
+						continue
```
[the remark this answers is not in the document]
blakerouse: I think it might be a small spike if the client is changing current status often during shutdown, but I think that will be minimum.

Really this is here to prevent the Elastic Agent to STOP then change its mind, there is no going back. Once Elastic Agent tells application to STOP it must stop and will not notify itself of any other configuration changes.

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+
+				var lastSent time.Time
+				var lastSentCfgIdx uint64
+				var lastSentStatus proto.StateObserved_Status
+				var lastSentMessage string
+				for {
+					select {
+					case <-done:
+						return
+					case <-time.After(500 * time.Millisecond):
```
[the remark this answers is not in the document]
blakerouse: I can make it configurable, but whats the need of doing that?

pkg/client/client.go
ph: @blakerouse sound good to me.

pkg/utils/utils.go
blakerouse: The `JSONMustMarshal` is only used for global error variables where the client sends back JSON back to the agent. So if they failed it would be on startup and during the unit tests.

pkg/client/client.go
blakerouse: Just a high enough number to handle the actions. I think 100 is really high in general, I think it could be even 10, but didn't want it to cause any blocking issues.

I don't think we should serialize the actions at the client level, I think if an action can conflict with another action then the application that implements the actions should handle that part. Just because an action request is received doesn't mean it must handle it immediately, if another action is prevent that action.

The actionsResponse channel is outside of the loop so responses will not be lost during reconnects.

pkg/client/client.go
blakerouse: Answered this above as well.

pkg/client/client.go
blakerouse: Yes, good idea! I have updated it to disconnect and reconnect.

pkg/client/client.go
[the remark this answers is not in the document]
blakerouse: Done

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+					}
+
+					c.cfgLock.RLock()
+					cfgIdx := c.cfgIdx
+					c.cfgLock.RUnlock()
+					c.obsLock.RLock()
```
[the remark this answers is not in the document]
blakerouse: Done.

pkg/client/client.go
```diff
@@ -0,0 +1,410 @@
...
+					sendMessage := func() {
+						err := checkinClient.Send(&proto.StateObserved{
+							Token:          c.token,
+							ConfigStateIdx: cfgIdx,
+							Status:         observed,
+							Message:        observedMsg,
+						})
+						if err != nil {
+							c.impl.OnError(err)
+							return
```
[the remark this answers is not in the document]
blakerouse: I have updated it to do just that.

pkg/client/client.go
blakerouse: Done

pkg/client/client.go
blakerouse: Good point, fixed.

pkg/client/client.go
blakerouse: I have moved them closer.

pkg/client/client.go
blakerouse: Added it both to `startCheckin` and `startActions`.

pkg/client/client.go
blakerouse: Fixed

pkg/client/client.go
blakerouse: Done

pkg/client/client.go
blakerouse: Fixed as well ;-)

PR #5, opened 2020-05-21
Add `client.NewFromReader` that creates a new client from reading bytes from the reader.

blakerouse:
Adds a new `client.NewFromReader` that creates a new client reading the bytes from the `io.Reader`. This will be used when passing the credentials to the spawned application, providing it with the information and credentials to connect back to the Elastic Agent GRPC server.

The packet of information is using protocol buffers as well allowing a single packet to be sent in most cases over stdin to the spawned application. Endpoint will get the same packet but through another socket.

The communication will use Mutual TLS so both the CA, Cert, and Key are sent to the application to connect back to the Agent along with the connection address and unique token.

**Example**:
```
client, err := client.NewFromReader(io.Stdin, clientImpl, actions...)
if err != nil {
    panic(err)
}
```


blakerouse:
This is actually dependent on #4, but github doesn't allow me to set it as a prerequisite. Once #4 is landed I will update the branch to make the diff correct.

blakerouse:
#4 is merged. This is now ready for a review now that the diff is not mixed up.

elastic-agent-client.proto
```diff
@@ -9,6 +9,22 @@ package proto;
 option cc_enable_arenas = true;
 option go_package = "pkg/proto;proto";
 
+// Connection information sent to the application on startup so it knows how to connected back to the Elastic Agent.
```
ruflin, line 12:
```suggestion
// Connection information sent to the process on startup so it knows how to connected back to the Elastic Agent.
```

Not sure if `process` is the correct one here, but we should align on one term.

pkg/client/client.go
```diff
@@ -89,12 +89,12 @@ type Client struct {
 	observed        proto.StateObserved_Status
 	observedMessage string
 
-	ctx    context.Context
-	cancel context.CancelFunc
-	wg     sync.WaitGroup
-	client proto.ElasticAgentClient
-	cfgLock   sync.RWMutex
-	obsLock   sync.RWMutex
+	ctx     context.Context
```
ruflin, line 92: We probably should have a formatter check as part of CI 🤔

pkg/client/reader_test.go
```diff
@@ -0,0 +1,230 @@
...
+	validClient, err := NewFromReader(connInfoReader, impl)
+	require.NoError(t, err)
+	require.NoError(t, validClient.Start(context.Background()))
+	defer validClient.Stop()
+	require.NoError(t, waitFor(func() error {
+		m.Lock()
+		defer m.Unlock()
+
+		if !gotValid {
+			return fmt.Errorf("server never recieved valid token")
```
ruflin, line 107:
```suggestion
			return fmt.Errorf("server never received valid token")
```

pkg/client/reader_test.go
```diff
@@ -0,0 +1,230 @@
...
+	connInfoBytes, err := protobuf.Marshal(connInfo)
+	require.NoError(t, err)
+	connInfoReader := bytes.NewReader(connInfoBytes)
+
+	impl := &StubClientImpl{}
+	validClient, err := NewFromReader(connInfoReader, impl)
+	require.NoError(t, err)
+	require.NoError(t, validClient.Start(context.Background()))
+	defer validClient.Stop()
+	require.NoError(t, waitFor(func() error {
```
ruflin, line 102: Where is the `waitFor` function defined?

pkg/client/reader_test.go
```diff
@@ -0,0 +1,230 @@
...
+	_, err := NewFromReader(bytes.NewBuffer([]byte("")), nil)
+	assert.Error(t, err)
+}
+
+func TestNewFromReader_BadProto(t *testing.T) {
+	_, err := NewFromReader(bytes.NewBuffer([]byte("bad proto")), nil)
+	assert.Error(t, err)
+}
+
+func TestNewFromReader_Connects(t *testing.T) {
```
ruflin, line 42: Nice test.

pkg/client/reader_test.go
```diff
@@ -0,0 +1,230 @@
...
+		},
+		NotBefore:             time.Now(),
+		NotAfter:              time.Now().AddDate(10, 0, 0),
+		IsCA:                  true,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageClientAuth, x509.ExtKeyUsageServerAuth},
+		KeyUsage:              x509.KeyUsageDigitalSignature | x509.KeyUsageCertSign,
+		BasicConstraintsValid: true,
+	}
+
+	privateKey, _ := rsa.GenerateKey(rand.Reader, 2048)
```
ruflin, line 139: I don't see a good reason an error should happen but probably still worth checking for it? If one happens, things for sure explode further down I guess.

pkg/client/client.go
ph: https://github.com/elastic/elastic-agent-client/issues/6 yup :)

elastic-agent-client.proto
blakerouse: `application` is the term used through out the protocol file

pkg/client/reader_test.go
blakerouse: Its defined in the client_test.go file that is in the same module

pkg/client/reader_test.go
blakerouse: Yeah it would explode because privateKey would be empty.

PR #10, opened 2020-06-05
proposal.init

michalpristas:
note: this is very early WIP
shareing just to create idea of usage


ph:
I think we can close it.

PR #13, opened 2020-06-11
Change pkg/client.Client to an interface

blakerouse:
Changes `New` and `NewFromReader` to return `Client` as an interface instead of a struct. This will allow others to mock the client in tests or other situations where it might be required.

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -62,7 +62,7 @@ pipeline {
         unstash 'source'
         insideGo{
           dir("${BASE_DIR}"){
-            sh(label: 'Unit test', script: 'go test ./... 2>&1 | tee test_results.txt')
+            sh(label: 'Unit test', script: 'go test -race ./... 2>&1 | tee test_results.txt')
             sh(label: 'Convert test results', script: 'go-junit-report < test_results.txt > junit.xml')
           }
         }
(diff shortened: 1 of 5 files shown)


ph:
@blakerouse seems we have found an issue with gRPC with the CI.

blakerouse:
@ph Strange, not able to reproduce locally.

ph:
@blakerouse Yes, this is strange indeed, I had the same reflex as you to update to the latest golang version.

@urso Have you seen this before https://beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Felastic-agent-client/detail/PR-13/4/tests ?

blakerouse:
Yeah still failed...

I don't get it.

urso:
> @urso Have you seen this before beats-ci.elastic.co/blue/organizations/jenkins/Beats%2Felastic-agent-client/detail/PR-13/4/tests ?

Not with GRPC, but I've seen something like this with unsafe pointers.

If `0xf` is the address that failed, then this might have been an 'offset' added to a null pointer.

ph:
Could that be a sync issue where value can be nil depending on the performance of the environment?

urso:
Might be. Have tried to compile with `-race` ?

ph:
@blakerouse This look like an issue with the test, lets try to make a followup for this.

blakerouse:
@ph Idk, seems like it was issue with go version as go 1.13 works fine.

Running 1.13 to align with beats locally, once beats is go 1.14 (waiting of vendor directory removal) then I will update the test back to go 1.14 and locally to see if I can fix the issue.

PR #15, opened 2020-07-09
Add new payload field to StateObserved

blakerouse:
Adds a new `payload` field to the `StateObserved` message. This will allow the client to send any extra information to the Elastic Agent when it checks in.


pkg/client/client.go
```diff
@@ -156,11 +157,21 @@ func (c *client) Stop() {
 }
 
 // Status updates the current status of the client in the Elastic Agent.
-func (c *client) Status(status proto.StateObserved_Status, message string) {
+func (c *client) Status(status proto.StateObserved_Status, message string, payload map[string]interface{}) error {
+	payloadStr := ""
+	if payload != nil {
+		payloadBytes, err := json.Marshal(payload)
```
michalpristas, line 163: payload is a map, will this result always in the same string?

PR #23, opened 2021-03-07
Extend the client to allow to register/unregister actions handlers

aleksmaus:
* Extend the Client interface with RegisterAction/UnregisterAction
* Update the signature of Execute API to pass the content, to allow cancelling with the context.Context if needed.

Needed to be able to register the actions with the gRPC client after it was created.


blakerouse (approved):
Looks good.

Issue #16, opened 2020-08-06
Leaking cancel contexts.


urso:
The client uses WithContext to create a context for signaling shutdown to sub go-routines. This seems to make the context leak when the workers return without cancelling the context. The cancel function must be run always. 

It seams the context that is generated on 'Start' is leaked at least if the Dial fails. The workaround is to call Stop always. There or other contexts that are generated via WithCancel, that look suspiciously like leaks, but I didn't check if that is the case.

ph:
@blakerouse @michalpristas  we probably have a leak?

michalpristas:
oh this one's sitting here for a while. we can plan it for this release, seems like easy fix
wdyt @ph

ph:
I will mark it for 7.11, the fix is easy but there are multiple PR involved.

PR #26, opened 2021-07-27
Leaks of connections and timers

michalpristas:
During debugging and profiling I noticed some small not really significant leaks but they appeared and it would be good to address them.

1st is timers usage where timer instance is kept even after loop iterations ends. I clean them after each loop iteration

2nd is wrapping connections with context with Cancel function so we close if client/server forgets to do so

3rd is proper closing when Send is done in grpc streams

### DIFF
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -120,11 +120,10 @@ type client struct {
 // New creates a client connection to Elastic Agent.
 func New(target string, token string, impl StateInterface, actions []Action, opts ...grpc.DialOption) Client {
 	actionMap := map[string]Action{}
-	if actions != nil {
-		for _, act := range actions {
-			actionMap[act.Name()] = act
-		}
+	for _, act := range actions {
+		actionMap[act.Name()] = act
 	}
+
 	return &client{
 		target:          target,
 		opts:            opts,
@@ -215,6 +214,7 @@ func (c *client) startCheckin() {
 			checkinCtx, checkinCancel := context.WithCancel(c.ctx)
 			checkinClient, err := c.client.Checkin(checkinCtx)
 			if err != nil {
+				checkinCancel()
 				c.impl.OnError(err)
 				continue
 			}
@@ -269,10 +269,14 @@ func (c *client) startCheckin() {
 				var lastSentMessage string
 				var lastSentPayload string
 				for {
+					t := time.NewTimer(500 * time.Millisecond)
 					select {
 					case <-done:
+						if !t.Stop() {
+							<-t.C
+						}
 						return
-					case <-time.After(500 * time.Millisecond):
+					case <-t.C:
 					}
 
 					c.cfgLock.RLock()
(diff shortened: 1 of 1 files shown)


pkg/client/client.go
```diff
@@ -215,6 +214,7 @@ func (c *client) startCheckin() {
 			checkinCtx, checkinCancel := context.WithCancel(c.ctx)
```
urso, line 214: it is safe to call cancel multiple times. But never forget to call cancel => use defer always. If you are inside a loop consider to factor out the body into a function. 

The function body is huge with many go-routines. Consider to split it up even further.

pkg/client/client.go
```diff
@@ -269,10 +269,14 @@ func (c *client) startCheckin() {
...
+					t := time.NewTimer(500 * time.Millisecond)
 					select {
 					case <-done:
+						if !t.Stop() {
+							<-t.C
+						}
 						return
-					case <-time.After(500 * time.Millisecond):
+					case <-t.C:
 					}
```
urso, line 280: This select statement is a common pattern. Replace with `timed.Wait(ctx, duration)` (timed package is part of go-concert).

michalpristas: this does not drain the channel now is designed to be used in select statements so i will skip this

pkg/client/client.go
```diff
@@ -366,6 +372,7 @@ func (c *client) startActions() {
 			actionsCtx, actionsCancel := context.WithCancel(c.ctx)
```
urso: same here, it should say `defer actionsCancel()`. The function body/loop requires some refactoring to ensure proper cleanup no matter what.

michalpristas: fixed

pkg/client/client.go
michalpristas: agreed offline that further refactoring of the funcs should be performed as a followup resulting in breaking one big function into smaller parts and getting rid of select statments in favor of proposed go-concert where possible

ruflin: ++ on follow up refactoring.

ruflin:
@michalpristas As a follow up, any idea on how we could have tests to cover these leaks to make sure they don't happen again?

michalpristas:
we could as part of e2e watch for number of fd maybe, but for this we would need longer run

urso:
> we could as part of e2e watch for number of fd maybe, but for this we would need longer run

Alternative consider to put process startup and management behind an interface. With proper interfaces and types it should be possible to actually write tests at the unit/integration testing level.

pkg/client/client.go
```diff
@@ -212,131 +211,139 @@ func (c *client) startCheckin() {
...
+					return
+				}
+				continue
+			}
 		}
 	}()
+
+	// wait for both send and recv go routines to stop before
+	// starting a new stream.
+	checkinWG.Wait()
```
urso, line 345: we have another helper in go-concert named `TaskGroup` ;)

urso (approved):

blakerouse (approved):
Changes look good. Thanks for tracking this down. #16

michalpristas:
@blakerouse i have couple of other fixes beats side so it fixes it partially

PR #29, opened 2022-03-15
V2 Control Protocol

blakerouse:
## V2 Control Protocol

This is a proposal on the new design of the control protocol.

### Features

* A unique state/config per unit (multiple inputs and output per process).
* A similar actions protocol (only added 1 extra field).
* A key-value storage scoped per unit for persistent storage.
* Access to the artifacts API through the control protocol.
* Ability to report name, version and metadata about the program to Elastic Agent.
* Ability to write log messages from processes that are not started as sub-processes.

### Backward Compatibility

This adds a new `CheckinV2` that is to be used when the Elastic Agent is operating in V2 mode. The original `Checkin` will be deprecated at some point in the future (once all tools have migrated off).

Closes https://github.com/elastic/elastic-agent/issues/215


blakerouse:
We need to decide on how we want to merge this into the repository. I was thinking about making it `github.com/elastic/elastic-agent-client/v2` keeping the current client as `github.com/elastic/elastic-agent-client`.

cmacknz:
> Use main for v2?

This will work but I've usually seen public proto packages versioned like Go packages, where v1 and v2 coexist together in the tree until one obsoletes the other. Like Go modules, this will make the import path change between versions. You could use `package proto;` and `package proto.v2;` for example and do the same on the generated Go modules.

Here is Google Cloud Storage using this approach: https://github.com/googleapis/googleapis/tree/master/google/storage

ph:
@cmacknz good point, but at the same time we have a limited set of users, so I am not sure having a v2 worth it. I believe this would reduce our flexibility to make changes in repo/testing and so on.

blakerouse:
I am +1 for using main for v2.

cmacknz:
> @cmacknz good point, but at the same time we have a limited set of users, so I am not sure having a v2 worth it. I believe this would reduce our flexibility to make changes in repo/testing and so on.

Fair point, go with whatever makes development easier.

elastic-agent-client.proto
```diff
@@ -9,86 +9,87 @@ package proto;
...
-    // Token that the application should send as the unique identifier when connecting over the GRPC.
-    string token = 3;
-    // CA certificate.
-    bytes ca_cert = 4;
-    // Peer certificate.
-    bytes peer_cert = 5;
-    // Peer private key.
-    bytes peer_key = 6;
+// Empty response message.
+message EmptyResponse {}
```
cmacknz, line 13: Is it worth using the `google.protobuf.Empty` well known type instead? https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Empty

It should be built into the proto compiler so it doesn't add a dependency.

elastic-agent-client.proto
```diff
@@ -112,15 +112,190 @@ message ActionResponse {
...
+    // Request and response is swapped here because the Elastic Agent sends the requests in a stream
+    // to the connected process. The order of response from the process does not matter, it is acceptable
+    // for the response order to be different then the request order.
     rpc Actions(stream ActionResponse) returns (stream ActionRequest);
+
+    // Key-Value state storage is provided for each unit.
+    //
+    // Transactional store is provided to allow multiple key operations to occur before a commit to ensure consistent
+    // state when multiple keys make up the state of an units persistent state.
+    rpc StoreBeginTxn(StoreBeginTxnRequest) returns (StoreBeginTxnResponse);
```
cmacknz, line 267: Is it worth defining the KV store as a separate `service`? You can have multiple services/clients on the same underlying gRPC connection.

Maybe only worth doing this if you'd want the ability to split the KV store into a process separate from the agent supervisor in the future.

ph: +1, Good comment here, Agree that would make it more future proofing.

elastic-agent-client.proto
```diff
@@ -9,86 +9,87 @@ package proto;
...
-    // Peer certificate.
-    bytes peer_cert = 5;
-    // Peer private key.
-    bytes peer_key = 6;
+// Empty response message.
+message EmptyResponse {}
+
+// State codes for the current state.
+enum State {
+    STARTING = 0;
```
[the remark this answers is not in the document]
ph: We will add state diagram to better define that.

ph:
@blakerouse The v1,v2  I am worried to run in a hybrid mode, wouldn't that bring more complexity while debugging?

elastic-agent-client.proto
blakerouse: If we want to split it, then we should probably split the artifacts, as that could move possibly in the future.

ph: You are right, lets split artifact too ++ thanks @cmacknz

jlind23:
@ph we will be aware when agent will be running in V2 mode, we just have to properly log it then.

blakerouse:
@ph Really we have 2 different ways we could implement it.

1. Run in v2 mode explicitly, so even though the `Checkin` function is present, we do not allow connection to it over the GRPC. Elastic Agent will error to the client that its in v2 mode and must connect over `CheckinV2`.
2. Run in hybrid mode that allows a v1 client but it translates to a v2 operating mode. So Elastic Agent runs in v2 mode, it computes the multiple units per process and builds the runtime environment. Then that v2 runtime environment is then merged to send a single configure to a v1 client, where the status is duplicated across the v2 runtime. This gives a temporary translation layer for v1 clients.

The open question is really do we want to support existing v1 clients when operating as v2, if so #2. Or we don't need too and #1 could be the answer. Or option #3, start with #1 and if we see an issue that requires us to do #2 then we implement the translation layer.

elastic-agent-client.proto
```diff
@@ -9,86 +9,87 @@ package proto;
...
+    HEALTHY = 2;
+    DEGRADED = 3;
+    FAILED = 4;
+    STOPPING = 5;
+}
+
+// Type of unit.
+enum UnitType {
+    INPUT = 0;
+    OUTPUT = 1;
```
ruflin, line 28: Output = shipper?

elastic-agent-client.proto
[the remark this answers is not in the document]
ruflin: Would be great to have a diagram of all the "objects" and dependencies. Would make it easier to an overview over the full protocol.

elastic-agent-client.proto
blakerouse: Not shipper, because this will be the GRPC to the shipper for the inputs and the actual output to elasticsearch, logstash, kafka, etc. in the shipper.

elastic-agent-client.proto
```diff
@@ -112,15 +112,190 @@ message ActionResponse {
...
+//
+// A started transaction must either have commit or discard called.
+message StoreBeginTxnRequest {
+    // Token that is used to unique identify the connection to the Elastic Agent.
+    string token = 1;
+    // ID of the unit.
+    string unit_id = 2;
+    // Type of transaction to start.
+    StoreTxnType type = 3;
+}
```
AndersonQ, line 131: Risking to sound a bit pedantic, but the comments use _start_ instead of _begin_, wouldn't be better to keep a consistent vocabulary? Therefore to change Begin on the names to Start?

elastic-agent-client.proto
AndersonQ, line 267: Question here: must all the store methods be called within a transaction, or could the Get|Set|Delete be called alone?

elastic-agent-client.proto
```diff
@@ -112,15 +112,190 @@ message ActionResponse {
...
+    // Key-Value state storage is provided for each unit.
+    //
+    // Transactional store is provided to allow multiple key operations to occur before a commit to ensure consistent
+    // state when multiple keys make up the state of an units persistent state.
+    rpc StoreBeginTxn(StoreBeginTxnRequest) returns (StoreBeginTxnResponse);
+    rpc StoreGetKey(StoreGetKeyRequest) returns (StoreGetKeyResponse);
+    rpc StoreSetKey(StoreSetKeyRequest) returns (EmptyResponse);
+    rpc StoreDeleteKey(StoreDeleteKeyRequest) returns (EmptyResponse);
+    rpc StoreCommit(StoreCommitRequest) returns (EmptyResponse);
+    rpc StoreDiscard(StoreDiscardRequest) returns (EmptyResponse);
```
AndersonQ, line 272: My GRPC is rusty, is there a default way to return errors?
If not, wouldn't the client be interested in knowing if these operations failed, so it must retry?

AndersonQ, line 272: For consistency, and also to avoid confusion, shouldn't those be suffixed with `Txn` as `Begin` is. Also just _discard_ can be interpreted as discarding a key, not a transaction. Commit not so much but still better to specify what is being committed.

elastic-agent-client.proto
```diff
@@ -9,86 +9,87 @@ package proto;
...
-// by the connected application.
+// A action request is streamed from the Elastic Agent to a unit so an action can be performed by that unit.
 message ActionRequest {
     // Unique ID of the action.
     string id = 1;
+    // Unique ID of the unit.
+    string unit_id = 2;
     // Name of the action.
-    string name = 2;
+    string name = 3;
```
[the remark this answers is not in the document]
cmacknz: The backwards compatibility rules are summarized here: https://developers.google.com/protocol-buffers/docs/proto3#updating

Adding a new enum entry is a backwards compatible change.

elastic-agent-client.proto
cmacknz: There is an implicit error returned by every RPC. In Go each of these rpc method will have an `error` return type. Really the return type is a gRPC status (https://pkg.go.dev/google.golang.org/grpc/status) encoded into an error.

elastic-agent-client.proto
```diff
@@ -112,15 +112,190 @@ message ActionResponse {
...
+
+    // Key-Value state storage is provided for each unit.
+    //
+    // Transactional store is provided to allow multiple key operations to occur before a commit to ensure consistent
+    // state when multiple keys make up the state of an units persistent state.
+    rpc StoreBeginTxn(StoreBeginTxnRequest) returns (StoreBeginTxnResponse);
+    rpc StoreGetKey(StoreGetKeyRequest) returns (StoreGetKeyResponse);
+    rpc StoreSetKey(StoreSetKeyRequest) returns (EmptyResponse);
+    rpc StoreDeleteKey(StoreDeleteKeyRequest) returns (EmptyResponse);
+    rpc StoreCommit(StoreCommitRequest) returns (EmptyResponse);
```
cmacknz, line 271: Note that if we decide in the future to return something other than an empty message, changing the return message type is not backwards compatible.

The way to guard against this is to define an empty `StoreCommitResponse` message as the return type, so that you can safely add fields to it in the future if you need to.

elastic-agent-client.proto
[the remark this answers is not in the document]
blakerouse: I disagree, an input can easily create any action they would like. By making this an enum that means the protocol needs to be updated every time 1 single input adds an action. We don't want that type of require on the protocol.

elastic-agent-client.proto
blakerouse: Sure I can update it.

elastic-agent-client.proto
blakerouse: Will add the `Txn` suffix.

elastic-agent-client.proto
blakerouse: Thats a good idea, will do that for all the other empty responses so in the future if return needs to be adjusted it can.

elastic-agent-client.proto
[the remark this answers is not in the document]
ph: I agree with @blakerouse, concerning the protocol. 

I have two proposal, to bring more strictness to the code:

- We could allow developers to define in the spec file (next to the binary) what actions they are supporting
- When we do the handshake we could receive a list of supported actions from the running process. This would not require changes on the protocol but would add more `strictness` here.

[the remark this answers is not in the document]
blakerouse: I think defining them in the spec file is a possibility but I don't know why it is really needed. Elastic Agent will only send an action that is targeted from Fleet to a specific input, for that action to occur in needs to be implemented in Fleet (or maybe as a package action in the future).

If we send an action to a process that doesn't understand that action it can just respond with an error, which is how the library that is in this repository already works. I think adding strictness even at the spec file level just seems like overhead that then the engineers will have to remember to add. To me the open question is why? I don't see a compelling reason.

blakerouse:
I have updated the protocol to taking into account all of the feedback. I have added back in the original `Checkin`, moving the new checkin too `CheckinV2`. I have also split out `ElasticAgentStore`, `ElasticAgentArtifacts` and `ElasticAgentMetrics`. I have also defined the parameters for the metrics protocol part.

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+    // Expected state of the unit.
+    State state = 3;
+    // Index of the either current configuration or new configuration provided.
+    uint64 config_state_idx = 4;
+    // Resulting configuration. (If the application already has the current `config_state_idx` this
+    // will be empty.)
+    string config = 5;
+}
+
+// A set of units and there expected states and configuration.
```
cmacknz, line 148:
```suggestion
// A set of units and their expected states and configuration.
```

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+// applied config index) along with the status of the application. In the case that the sent `config_state_idx`
+// doesn't match the expected `config_state_idx` that Elastic Agent expects, the unit is always marked as
+// `CONFIGURING` and a new `UnitExpected` will be sent to so it can have the latest configuration.
+message UnitObserved {
+    // Unique ID of the unit.
+    string id = 1;
+    // Unit type.
+    UnitType type = 2;
+    // Current index of the applied configuration.
+    uint64 config_state_idx = 3;
```
cmacknz, line 165: I don't understand what this parameter does. What is it an index into? Is it an identifier?

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+    // State of unit.
+    State status = 4;
+    // Message for the health status.
+    string message = 5;
```
cmacknz, line 169: Is this just a message providing more detail on the status? Like an error message?

Where is it displayed? Logs or is it user visible?

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+// Response of the retrieved key.
+message StoreGetKeyResponse {
+    // Status result of the get.
+    enum Status {
+        // Action was successful.
+        FOUND = 0;
+        // Action has failed.
+        NOT_FOUND = 1;
+    }
+    Status status = 1;
```
cmacknz, line 252: Could you just communicate this information using the error returned by the RPC call?

There are built in status codes you could use for this. `OK` means found, `NOT_FOUND` means not found: https://grpc.github.io/grpc/core/md_doc_statuscodes.html

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+}
+
+// Response from `MetricsReport`.
+message MetricsReportResponse {
+    // Empty at the moment, defined for possibility of adding future return values.
+}
+
+service ElasticAgentMetrics {
+    // Report metrics to the Elastic Agent.
+    rpc Report(MetricsReportRequest) returns (MetricsReportResponse);
```
cmacknz, line 413: Where do these metrics go? I am curious how you avoid re-implementing parts of metricbeat or the data plane in general in the long term.

There was some consideration given to adding a generalized metric endpoint to the shipper, supporting an OpenTelemetry metrics or Statsd endpoint or similar there would have value for the future metricbeat+integrations beyond just monitoring agent processes.

elastic-agent-client.proto
ph: This would be used visible, logs or even in fleet.

elastic-agent-client.proto
ph: I have mixed feelings on this, I believe we could use the predefined code. For a locally K/V store that would make sense. Would these return code be OK or prefixed enough if we move to an external K/V Storage like Elasticsearch?

cmacknz: The errors from RPC calls in Go will have both a code and a string error message, see https://jbrandhorst.com/post/grpc-errors/ for examples.

The point is to allow easily handling the standardized codes as standard errors in clients (like HTTP status codes), but providing an error message to add context.

I think the codes are doing the same job as your enum here. They are widely used in the gRPC ecosystem.

elastic-agent-client.proto
cmacknz: Can that be documented here in the protocol so it's obvious how it will be used?

elastic-agent-client.proto
ph: @cmacknz @blakerouse Let's have a brainstorm on this to to make sure we are aligned on the goal.  I am not too worried about reimplementing the data plane, I would consider these metrics, agent health information and not something that metricbeat would collect.

elastic-agent-client.proto
ph: You convince me, here I don't have an argument agains't it.

michel-laterman:
Just to get wider feedback for some diagnostics concerns i've talked to @blakerouse about:

1. We need a way for units to be able to identify themselves either on startup, or on demand. This information will include things like the version number as well as build information such as git commit, build time, etc. It should also include any ids that are used by the agent (such as any ephemeral ids).
2. There needs to be an easy way to determine what config a unit is running. It can either be tracked by the agent or retrieved by an rpc call.
3. We need a way for the agent to be able to gather diagnostics information from a unit. Diagnostics info can be arbitrarily defined by a unit (currently we use pprof info from go processes). A call for diagnostics info can create multiple return values (such as pprof data for trace, heap, etc).

elastic-agent-client.proto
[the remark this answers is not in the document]
kvch: It seems to me that this list is the common states of both inputs and outputs. However, I would like to provide a more granular state reporting for inputs. For example, I would like to report if an input is actively publishing events or it is idle. Where should those states go? Are you open to extend this list with unit specific states?

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+    // Transaction ID.
+    string txn_id = 2;
+}
+
+// Response from `DiscardTxn`.
+message StoreDiscardTxnResponse {
+    // Empty at the moment, defined for possibility of adding future return values.
+}
+
+service ElasticAgentStore {
```
kvch, line 329: Why is this store needed? Who are the users of this?

elastic-agent-client.proto
blakerouse: It is the currently supplied index of the configuration. So the Elastic Agent doesn't need to send the full configuration to the process, it sends the index of the configuration that the process should be on. The process also responds with its index as well. If there is a mismatch the Elastic Agent will send a new observed message with the correct index and the configuration content. Any time the configuration changes it increments the index, so the program knows that it needs to receive the latest configuration.

elastic-agent-client.proto
blakerouse: This is for all users, to ease their implementation. Hopefully filestreams and other inputs will transition to using the key-value store. The benefit is the life-cycle of the state is handled by the Elastic Agent.

Elastic Agent will clean up state for units, as well manage the state across upgrade of versions. No need for the inputs to worry about writing state and managing the life-cycle of it.

elastic-agent-client.proto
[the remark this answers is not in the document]
blakerouse: @kvch I have a feeling that will result in a lot of flipped states from publish to not published. All of these state changes will be reported to Fleet so it can be exposed and viewable. I would think that providing something that granular and the fact that it would probably flip back and forth a lot this is not the place we want to do that.

I would take it as a healthy input is one that will report events when they exist. So if the input cannot even read a directory, or if it cannot connect to AWS then it should not be reported as healthy.

elastic-agent-client.proto
```diff
@@ -9,86 +9,87 @@ package proto;
...
+    // Unique ID of the unit.
+    string id = 1;
+    // Unit type.
+    UnitType type = 2;
     // Current index of the applied configuration.
     uint64 config_state_idx = 2;
-    // Status code.
-    Status status = 3;
+    // State of unit.
+    State status = 3;
```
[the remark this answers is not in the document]
blakerouse: Comes from v1, cannot be changed. Was fixed in v2.

elastic-agent-client.proto
blakerouse: Added.

elastic-agent-client.proto
```diff
@@ -112,15 +114,301 @@ message ActionResponse {
...
+    // ID of the artifact.
+    string id = 2;
+    // SHA256 of the artifact.
+    string sha256 = 3;
+}
+
+// Content of the artifact.
+message ArtifactFetchResponse {
+    // Artifact content.
+    bytes content = 1;
```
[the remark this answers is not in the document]
blakerouse: Changed

blakerouse:
Changes in the latest iteration:

```
* Ability to report name, version and metadata about the program to Elastic Agent.
* Ability to write log messages from processes that are not started as sub-processes.
```

Other small comment suggestions and the response body for artifacts service.

jlind23:
@blakerouse are we good to go with a first version of this draft or is there still some unanswered comments that you need to address?

elastic-agent-client.proto
```diff
@@ -112,15 +115,361 @@ message ActionResponse {
...
+// A unit that is part of a collector/shipper.
+message UnitExpected {
+    // Unique ID of the unit.
+    string id = 1;
+    // Unit type.
+    UnitType type = 2;
+    // Expected state of the unit.
+    State state = 3;
+    // Index of the either current configuration or new configuration provided.
+    uint64 config_state_idx = 4;
```
AndersonQ, line 143: I'm a bit puzzled by the idea that the meaning of this field changes... If this type represents what is expected of a unit, then shouldn't `config_state_idx` always be the config index it should be on?
If the Agent knows the unit is already on the expected `config_state_idx` it won't populate `config`, fine. However it should not change the meaning of `config_state_idx`. It's always the the config index the unit should be on

elastic-agent-client.proto
```diff
@@ -112,15 +115,361 @@ message ActionResponse {
...
+message UnitObserved {
+    // Unique ID of the unit.
+    string id = 1;
+    // Unit type.
+    UnitType type = 2;
+    // Current index of the applied configuration.
+    uint64 config_state_idx = 3;
+    // State of unit.
+    State status = 4;
+    // Nice message for the health status of the unit.
```
AndersonQ, line 169: [NIT]
Perhaps something like the following is more clear and less abstract.
```suggestion
    // Human readable message for the health status of the unit.
```

elastic-agent-client.proto
```diff
@@ -112,15 +115,361 @@ message ActionResponse {
...
+    // Exposed to users to provide more detail about the status for this single unit.
+    string message = 5;
+    // JSON encoded payload for the status.
+    string payload = 6;
```
AndersonQ, line 173: Any special reason for having here a JSON as string whereas on the other places it's bytes?

elastic-agent-client.proto
```diff
@@ -112,15 +115,361 @@ message ActionResponse {
...
+    repeated UnitMetric metrics = 2;
+}
+
+message MetricsReportRequest {
+    // Token that is used to uniquely identify the collection of inputs to the agent. When started this is provided
+    // in the `ConnInfo`.
+    string token = 1;
+
+    // Period of time for the metrics that are collected.
+    uint64 period = 2;
```
AndersonQ, line 417: I believe the time unit should be defined/stated here. Otherwise just a number is unclear

elastic-agent-client.proto
```diff
@@ -112,15 +115,361 @@ message ActionResponse {
...
+// This allows services that are not started as sub-processes to write to the same stdout that programs that do get
+// started as subprocess to write to. A program that is as a sub-process with stdout connected does not have the
+// ability to use this service.
+service ElasticAgentLog {
+    // Log a message to the Elastic Agent.
+    rpc Log(LogMessageRequest) returns (LogMessageResponse);
+
+    // Bulk log messages to the Elastic Agent.
+    rpc LogBulk(LogBulkMessageRequest) returns (LogBulkMessageResponse);
+}
```
AndersonQ, line 475: Do we really need 2 APIs for that? One bulk API where it might have one or more messages isn't enough?

elastic-agent-client.proto
```diff
@@ -112,15 +115,361 @@ message ActionResponse {
...
+
+message LogBulkMessageResponse {
+    // Empty at the moment, defined for possibility of adding future return values.
+}
+
+// Log service is only exposed to programs that are not started as sub-processes by Elastic Agent.
+//
+// This allows services that are not started as sub-processes to write to the same stdout that programs that do get
+// started as subprocess to write to. A program that is as a sub-process with stdout connected does not have the
+// ability to use this service.
```
AndersonQ, line 468: I found it a bit odd... the program reporting logs isn't writing to a standard output. It's reporting logs, perhaps reporting what they're writing to stdout so the Agent can monitor them. Besides I believe it should be both stdout and stderr

elastic-agent-client.proto
ph: Agree on this one, let's just use a `bulk`

elastic-agent-client.proto
ph: @AndersonQ This is only for endpoint, we aren't controlling the process it's started and controller by the operating system, so we cannot in that specific case monitor the stdout or stderr.

ph:
@jlind23 I am good for this as first version, there are two point we need to clarify which is diagnostic and metrics, we have a call today/tomorrow. but I think we can do it as an increment of this version.

elastic-agent-client.proto
blakerouse: What it means is the expected configuration index for that unit. So in the case its the same the unit already has the idx will be the same value as the previous with the contents of config missing. In the case that it needs to update the configuration its idx++ with the configuration provided.

elastic-agent-client.proto
blakerouse: Updated.

blakerouse:
@jlind23 I think its really close. Just need to wrap up the 2 things for diagnostics and metrics.

blakerouse:
@michel-laterman Added the diagnostic action and defined response messages. Give it a look over, make sure I captured everything.

michel-laterman:
Yep, think everything we need is there.

@ferullo, what kind of information do you think endpoint will report as far as diagnostics goes? Do we need to add anything or is the current message structs generic enough to work?

ph:
fyi @jlind23 

# Next Steps
- Remove metrics as a goal of this PR
- Write a proposal for metrics and logging (state manamm)

---

# Sync over zoom
Attendees (@blakerouse, @cmacknz , @ph)

## Goals
- We don't want to depend on Fleet or Elasticsearch to get the monitoring data we need for SDH. 
- We want to it to be part of a diagnostic
- Be realtime/live (tail -f)
- I want to be able to take the data from a diagnostic and graph it.
- We should always collect metrics (do not turn them off)
- Do not use metricbeat

## How long Elastic Agent should keep that information
- We need 30sec and 5 mins resolution
-  Memory is enough

## OpenTelemetry
- TODO Ask APM what they think about it.
- How they would do a simplier or lightweight version of APM.
- Need to be aggregated by the Elastic Agent

## How we ship metric that data to elasticsearch
- Throught the shipper?
- Though fleet server? This one
- APM-Server

## APM Tracing
- Not enable all the time because of the desktop machine
- When we turn it on, we start the necessary components to send the APM data to Elasticsearch
- Start a second shipper so the traces aren't mixed.

## Are we sending Logging data via Fleet Server
- Observabable
  - CPU usage
  - memory usage
- The unit can reports metrics informations
  -  Input Provides how many events per sec/protocol
  -  Shippers Provides metrics receives / sends
  - How many files handlers

## Reference
https://stripe.com/blog/canonical-log-lines
https://github.com/elastic/elastic-agent-shipper/issues/11

## The Questions
- Is Fleet status is the **current** state in time or **overtime** status is from the log statement collected

blakerouse:
I have removed the metrics service from the protocol as discussed. I am now generating the protocol code and working on the client to enable v2.

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+    // Response from this call can be chunked over multiple `ArtifactFetchResponse` for very large responses. A minimum
+    // of two responses will always be returned. The last response has eof set.
+    rpc Fetch(ArtifactFetchRequest) returns (stream ArtifactFetchResponse);
+}
+
+message LogMessage {
+    // ID of the unit.
+    string unit_id = 1;
+    // ECS log message body JSON encoded.
+    bytes message = 2;
```
ruflin, line 417: Should we enforce some basic structure already here. For example having timestamp and message already part of the schema and only the rest of the info a blob? This ensures that at least the basic log message is valid.

ph: I am unsure here, should we assume that the application uses an ECS format appender to the logs?

blakerouse: I think any enforcement here will add overhead. The idea here is that its transparently takes the bytes and just outputs it through stdout.

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+    // state when multiple keys make up the state of an units persistent state.
+    rpc BeginTxn(StoreBeginTxnRequest) returns (StoreBeginTxnResponse);
+    rpc GetKey(StoreGetKeyRequest) returns (StoreGetKeyResponse);
+    rpc SetKey(StoreSetKeyRequest) returns (StoreSetKeyResponse);
+    rpc DeleteKey(StoreDeleteKeyRequest) returns (StoreDeleteKeyResponse);
+    rpc CommitTxn(StoreCommitRequest) returns (StoreCommitTxnResponse);
+    rpc DiscardTxn(StoreDiscardRequest) returns (StoreDiscardTxnResponse);
+}
+
+// Requests an artifact from the Elastic Agent.
```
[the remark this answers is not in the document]
blakerouse: It is meant as a replacement for Endpoint downloading directly from Fleet Server. We want to remove that and have artifact access to go through the control protocol.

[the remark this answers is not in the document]
ph: @ferullo it's an evolution, we aren't planning to remove it in the short term from fleet-server, there were also needs for inputs in beats world to download larger assets and we would like to have a more uniform way of doing it, so beats don't reach out directly to the fleet server, this also aligns with our goal to lock down more the beats.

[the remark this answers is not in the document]
ferullo: 👍

@kevinlog can you think of anything more that would be needed here for Endpoint's use case?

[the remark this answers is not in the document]
kevinlog: We are currently POC'ing a feature to request and download files from the Endpoint for `8.4`.  The user would be able to specify a path on the remote host and the Endpoint would call some function to transfer the file to the server.

We were thinking about putting some logic into Fleet Server to help with storing the files via artifacts (and eventually ES).  I don't think this precludes us from doing this, but it's good to have the conversation.

@pzl is working on a POC for this ticket, currently: https://github.com/elastic/security-team/issues/3616. @pzl - any thoughts on the above?  Does this change things?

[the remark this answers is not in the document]
pzl: As mentioned above, we're (Endpoint) starting to map out the approach for bi-directional file transfers. Unclear as of yet if it will be using the same "artifacts" structure in use right now that Endpoint fetches.

Am I understanding correctly that this is an iteration on Artifact delivery, such that the client<->Agent protocol now assists in the delivery of those artifacts? Is the plan to move artifacts off of fleet server? Would agent be fetching them from Kibana?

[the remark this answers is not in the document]
blakerouse: The Elastic Agent never talks to Kibana. This is changing the behavior so endpoint does not need to be passed credentials to communicate with Fleet Server. Instead the Elastic Agent will perform the talking to Fleet Server. Only the Elastic Agent should communicate with Fleet Server, everything underneath in the control protocol sense should go through the Elastic Agent on the system.

Nothing actually changes on Fleet Server side, the change is that in v2 Endpoint will not be given the credentials or the URL to even communicate with Fleet Server. Endpoint will request the artifact from Elastic Agent and then Elastic Agent will get that artifact from the Fleet Server and provide it to Endpoint through the GRPC control protocol.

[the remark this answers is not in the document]
pzl: Ah, thank you. I think I understand the shifting responsibilities now.

Where this does conflict a little bit with our future plans is that we were looking to _expand_ some of the fleet APIs to handle this bi-directional file handling. And we were looking to have Endpoint fetch and deliver more files outside of the normal Artifact distribution. (We would be putting resources into adding these API changes to fleet server)

If the future-looking plans are to nix the Fleet Server -> client (Endpoint) communications and credentials, then perhaps we need to rethink. Is this something we can include in the V2 protocol here?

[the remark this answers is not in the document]
blakerouse: Yes we can add it to the v2 protocol. Let me know what you need, and we can align to ensure that we provide it for you.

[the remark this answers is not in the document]
ph: @kevinlog One question about uploading any reason that the endpoint doesn't send that artifact directly to an Elasticsearch index? What is the rational to pass through fleet server?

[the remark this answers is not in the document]
pzl: For large files (dozens to hundreds of MB, perhaps into GB range), it should be chunked into many documents to my understanding. Endpoint could potentially perform this chunking work on its own and send into ES indices/streams:

but for future-looking scenarios where we have better storage options (either an object store like was recently made available in our cloud deployments, but not yet usable by solutions, or if ES adds native file-storage https://github.com/elastic/dev/issues/1544 ) we don't want to expose the underlying storage mechanism so that the API can remain stable and the storage solution can evolve. The other problem is one of access control and permissions for the Endpiont, which is handled more easily through a fleet server API

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+    // Key-Value state storage is provided for each unit.
+    //
+    // Transactional store is provided to allow multiple key operations to occur before a commit to ensure consistent
+    // state when multiple keys make up the state of an units persistent state.
+    rpc BeginTxn(StoreBeginTxnRequest) returns (StoreBeginTxnResponse);
+    rpc GetKey(StoreGetKeyRequest) returns (StoreGetKeyResponse);
+    rpc SetKey(StoreSetKeyRequest) returns (StoreSetKeyResponse);
+    rpc DeleteKey(StoreDeleteKeyRequest) returns (StoreDeleteKeyResponse);
+    rpc CommitTxn(StoreCommitRequest) returns (StoreCommitTxnResponse);
+    rpc DiscardTxn(StoreDiscardRequest) returns (StoreDiscardTxnResponse);
```
AndersonQ, line 381: [Consistency]
Same as above
```suggestion
    rpc CommitTxn(StoreCommitTxnRequest) returns (StoreCommitTxnResponse);
    rpc DiscardTxn(StoreDiscardTxnRequest) returns (StoreDiscardTxnResponse);
```

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+
+// Requests an artifact from the Elastic Agent.
+message ArtifactFetchRequest {
+    // Token that is used to uniquely identify the collection of inputs to the agent. When started this is provided
+    // in the `ConnInfo`.
+    string token = 1;
+    // ID of the artifact.
+    string id = 2;
+    // SHA256 of the artifact.
+    string sha256 = 3;
```
AndersonQ, line 392: How would the client already know the SHA256 of the artifact? During the build we'd embed all valid artifact IDs and SHAs?

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+
+message LogMessageResponse {
+    // Empty at the moment, defined for possibility of adding future return values.
+}
+
+// Log service is only exposed to programs that are not started as sub-processes by Elastic Agent.
+//
+// This allows services that are not started as sub-processes to write to the same stdout that programs that do get
+// started as subprocess to write to. A program that is as a sub-process with stdout connected does not have the
+// ability to use this service.
```
AndersonQ, line 435: Typo?
```suggestion
// Log service is only exposed to programs that are not started as sub-processes by Elastic Agent.
//
// This allows services that are not started as sub-processes to write to the same stdout that programs that do get
// started as subprocess write to. A program that is as a sub-process with stdout connected does not have the
// ability to use this service.
```

elastic-agent-client.proto
AndersonQ: > I think any enforcement here will add overhead.

+1

So would the agent log it in a similar manner to what id does today, wrapping the application's log message on the Agent's own log message? Similar to:

```
message: "Application X log: Whole application's log"
```

AndersonQ (changes requested):

ph (approved):
I though I did approve this before.
I think @AndersonQ comments about the naming are appropriate, we can address them and start iterating.

elastic-agent-client.proto
AndersonQ: Let me try to rephrase it:
Do the agent just logs whatever it receives "as is", as plains as `fmt.Println(message)` or does it do or add anything before logging?

ph: I think we should consider the log ready to ready or "as is" and don't do any processing on it.

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
     string token = 1;
     // Unique ID of the action.
     string id = 2;
     // Status of the action.
     Status status = 3;
-    // JSON encoded result for the action.
+    // JSON encoded result for the action (empty when diagnostic action response).
     bytes result = 4;
+    // Specific result for diagnostics action. (only used in V2)
+    repeated ActionDiagnosticUnitResult diagnostic = 5;
```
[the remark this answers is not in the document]
blakerouse: Yeah type is not needed in the result, because we have the ID of the action. The type is taken by looking at the request. We would not care what the response says the type is, we would never trust that.

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+    STOPPING = 5;
+}
+
+// Type of unit.
+enum UnitType {
+    INPUT = 0;
+    OUTPUT = 1;
+}
+
+// A unit that is part of a collector/shipper.
```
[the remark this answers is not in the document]
blakerouse: There is no requirement saying that it can only have 1. In the future might be possible to have a shipper that sends the same event to 2 different outputs.

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+// In the case that a unit is missing from the observation then the Elastic Agent will mark that missing unit
+// as `STARTING` and send a new `UnitExpected` for the missing unit.
+message CheckinObserved {
+    // Token that is used to uniquely identify the connection to the Elastic Agent.
+    string token = 1;
+    // Units observed state.
+    repeated UnitObserved units = 2;
+    // Version information about the running program. Should always be included on first checkin, and not again unless
+    // one of the values have changed.
+    optional CheckinObservedVersionInfo version_info = 3;
```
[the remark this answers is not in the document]
blakerouse: No this is the program that is executed by Elastic Agent version information, not Elastic Agents own information.

elastic-agent-client.proto
```diff
@@ -101,26 +129,311 @@ message ActionResponse {
...
+    // Token that is used to uniquely identify the connection to the Elastic Agent.
+    string token = 1;
+    // Transaction ID.
+    string txn_id = 2;
+}
+
+// Response from `DiscardTxn`.
+message StoreDiscardTxnResponse {
+    // Empty at the moment, defined for possibility of adding future return values.
+}
```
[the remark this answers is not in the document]
blakerouse: Updated.

elastic-agent-client.proto
blakerouse: This is how the protocol works. Best to look up how the Fleet Server artifact API works, don't want to conflate this PR with that information.

jlind23:
@ph @blakerouse are we good to make it move forward?
cc @pierrehilbert

elastic-agent-client.proto
```diff
@@ -101,26 +147,312 @@ message ActionResponse {
...
+
+service ElasticAgentStore {
+    // Key-Value state storage is provided for each unit.
+    //
+    // Transactional store is provided to allow multiple key operations to occur before a commit to ensure consistent
+    // state when multiple keys make up the state of an units persistent state.
+    rpc BeginTxn(StoreBeginTxnRequest) returns (StoreBeginTxnResponse);
```
aleksmaus, line 393: nit: maybe rename BeginTxn -> BeginTx and other Txns to Txs?
similar to postgres go library for example BeginTx
https://github.com/lib/pq/blob/1ef134dc0e0dcdf8097ca347580c15df72ef786e/conn_go18.go#L58
or mysql
https://github.com/go-sql-driver/mysql/blob/bcc459a906419e2890a50fc2c99ea6dd927a88f2/connection.go#L471

"Tx" is more common abbreviation for transactions in other db libaries

elastic-agent-client.proto
```diff
@@ -101,26 +147,312 @@ message ActionResponse {
...
+    // Value of the key.
+    bytes value = 4;
+    // TTL of the key (in milliseconds)
+    uint64 ttl = 5;
+}
+
+// Response from `SetKey`.
+message StoreSetKeyResponse {
+    // Empty at the moment, defined for possibility of adding future return values.
+}
```
ph, line 339: Good catch, I see a TTL maybe or single-use key.

elastic-agent-client.proto
```diff
@@ -101,26 +147,312 @@ message ActionResponse {
...
+
+// Deletes a key in the store.
+//
+// `txn_id` must be an ID of a transaction that was started with `READ_WRITE`.
+//
+// Does not error in the case that a key does not exist.
```
ph, line 345: +1, align with Go's expected behavior.

pzl:
@blakerouse @ph 

What's the procedure look like for additions to this protocol in the future? If the changes are purely additive to the control protocol, can things be added to this spec in, say, 8.4 timeline?  

We are still looking to add a new feature/API to Fleet Server for our next use case, but we don't have a full picture of what's being sent over the wire (aside from, its a big file. Not sure if streamed or pre-chunked. Metadata format, etc)

I assume you want this merged without waiting for us? Or is it much preferred to hammer out something on our side and send it in here

elastic-agent-client.proto
aleksmaus: depending on the storage backend/library and transaction implementation details, not closing the transaction could leave the transaction locks on the database/tables/rows. should probably be coded to have transaction in the scope of the connection or something and the connection is broken the transaction should be auto rolled back. thinking if these  kind of cases. don't completely know if this is relevant for our particular implementation.

elastic-agent-client.proto
kvch: On the surface-level it sounds great. But after thinking about it, I am on the fence. If e.g. `filestream` uses agent to store state information, it means that we cannot really test `filestream` without either starting an Agent or mocking it. I am afraid this adds unnecessary overhead to development of inputs. During testing I would like to avoid starting an Agent whenever it is possible. In e2e tests it is fine, but for unit and integration tests it sounds too much. But let's see.

pkg/client/client_v2.go
```diff
@@ -0,0 +1,541 @@
...
+	done := make(chan bool)
+
+	// expected state check-ins
+	checkinRead.Add(1)
+	go func() {
+		defer checkinRead.Done()
+		for {
+			expected, err := checkinClient.Recv()
+			if err != nil {
+				if err != io.EOF {
```
ph, line 200: This would fail the linter.

```suggestion
				if !errors.Is(err, io.EOF) {

File: pkg/client/client_v2.go [hunk 36]
```diff
@@ -0,0 +1,541 @@
...
+		defer checkinWrite.Done()
+
+		if err := c.sendObserved(checkinClient); err != nil {
+			if err != io.EOF {
+				c.errChan <- err
+			}
+			return
+		}
+
+		t := time.NewTicker(c.minCheckTimeout)
```
ph, line 222: Based on our morning discussion this would behavior weirdlky with sleep or hibernate?

pkg/client/client_v2.go
```diff
@@ -0,0 +1,541 @@
...
+			c.units[i] = unit
+			i++
+		} else {
+			c.unitsChan <- UnitChanged{
+				Type: UnitChangedRemoved,
+				Unit: unit,
+			}
+		}
+	}
+	c.units = c.units[:i]
```
ph, line 296: I've almost missed the resize.

pkg/client/client_v2.go
```diff
@@ -0,0 +1,541 @@
...
+			case proto.ActionRequest_CUSTOM:
+				c.tryPerformAction(actionResults, action)
+			case proto.ActionRequest_DIAGNOSTICS:
+				// TODO: Implement the diagnostics action.
+				// At the moment it just returns action type unknown until implemented.
+				actionResults <- &proto.ActionResponse{
+					Token:  c.token,
+					Id:     action.Id,
+					Status: proto.ActionResponse_FAILED,
+					Result: ActionTypeUnknown,
```
ph, line 402: +1

pkg/client/client_v2.go
```diff
@@ -0,0 +1,541 @@
...
+	c.unitsLock.RUnlock()
+	if unit == nil {
+		actionResults <- &proto.ActionResponse{
+			Token:  c.token,
+			Id:     action.Id,
+			Status: proto.ActionResponse_FAILED,
+			Result: ActionErrUnitNotFound,
+		}
+		return
+	}
```
ph, line 470: So this mean that If an agent receive an `osquery` and it receive an osquery action and the unit isn't runnning we would silently ignore the event?

pkg/client/client_v2.go
```diff
@@ -0,0 +1,541 @@
...
+	err := json.Unmarshal(action.Params, &params)
+	if err != nil {
+		actionResults <- &proto.ActionResponse{
+			Token:  c.token,
+			Id:     action.Id,
+			Status: proto.ActionResponse_FAILED,
+			Result: ActionErrUnmarshableParams,
+		}
+		return
+	}
```
ph, line 497: thanks for adding this, +1 don't trust anything.

pkg/client/client_v2_test.go
```diff
@@ -0,0 +1,696 @@
...
+						"custom": "payload",
+					})
+				case UnitChangedModified:
+					state, _ := change.Unit.Expected()
+					if state == UnitStateStopped {
+						change.Unit.UpdateState(UnitStateStopping, "Stopping", nil)
+						go func() {
+							time.Sleep(100 * time.Millisecond)
+							change.Unit.UpdateState(UnitStateStopped, "Stopped", nil)
+						}()
```
ph, line 286: Should we make the "Healthy", "Stopping", "Stopped" and other a public constant?

pkg/client/client_v2_test.go
```diff
@@ -0,0 +1,696 @@
...
+	}()
+
+	require.NoError(t, client.Start(ctx))
+	defer client.Stop()
+	require.NoError(t, waitFor(func() error {
+		m.Lock()
+		defer m.Unlock()
+
+		if !connected {
+			return fmt.Errorf("server never recieved valid token")
```
ph, line 305:
```suggestion
			return fmt.Errorf("server never received a valid token")

File: pkg/client/client_v2_test.go [hunk 43]
```diff
@@ -0,0 +1,696 @@
...
+			case <-ctx.Done():
+				return
+			case change := <-client.UnitChanges():
+				switch change.Type {
+				case UnitChangedAdded:
+					unitsLock.Lock()
+					units = append(units, change.Unit)
+					unitsLock.Unlock()
+				default:
+					panic("not implemented")
```
ph, line 389: let's use testing's `t.Fatal`

pkg/client/client_v2_test.go
```diff
@@ -0,0 +1,696 @@
...
+	}()
+
+	require.NoError(t, client.Start(ctx))
+	defer client.Stop()
+	require.NoError(t, waitFor(func() error {
+		m.Lock()
+		defer m.Unlock()
+
+		if !gotInit {
+			return fmt.Errorf("server never recieved valid token")
```
ph, line 402:
```suggestion
			return fmt.Errorf("server never received valid token")

File: pkg/client/common.go [hunk 45]
```diff
@@ -0,0 +1,57 @@
...
+
+package client
+
+import (
+	"time"
+
+	"github.com/elastic/elastic-agent-client/v7/pkg/utils"
+)
+
+// CheckinMinimumTimeout is the amount of time the client must send a new checkin even if the status has not changed.
```
ph, line 13: See above discussion concerning hibernation and sleep.

pkg/client/common.go
```diff
@@ -0,0 +1,57 @@
...
+// when the request action unit cannot be found.
+var ActionErrUnitNotFound = utils.JSONMustMarshal(map[string]string{
+	"error": "action unit not found",
+})
+
+// ActionTypeUnknown is returned to Elastic Agent as result to an action request
+// where the action type is unknown.
+var ActionTypeUnknown = utils.JSONMustMarshal(map[string]string{
+	"error": "action type unknown",
+})
```
ph, line 57: We should probably have a concrete type implementing `MarshalJSON` instead of creating JSON fragment, something like:

```golang
ActionTypeErr{ error: "action-type unknown")
```

pkg/client/store_test.go
```diff
@@ -0,0 +1,496 @@
...
+				return
+			case change := <-client.UnitChanges():
+				switch change.Type {
+				case UnitChangedAdded:
+					unitsLock.Lock()
+					units = append(units, change.Unit)
+					unitsLock.Unlock()
+				default:
+					panic("not implemented")
+				}
```
ph, line 74: `t.Fatal`

pkg/client/store_test.go
```diff
@@ -0,0 +1,496 @@
...
+		unitStore = &unitMemoryStore{
+			unitID:   unitID,
+			unitType: unitType,
+			data:     make(map[string]unitMemoryValue),
+			txns:     make(map[string]*unitMemoryStoreTxn),
+		}
+		s.store = append(s.store, unitStore)
+	}
+	return unitStore
+}
```
ph, line 496: Well, we have a full in-memory store :D

pkg/client/unit.go
```diff
@@ -0,0 +1,218 @@
...
+// Unit represents a distinct item that needs to be operating with-in this process.
+//
+// This is normally N number of inputs and 1 output (possible for multiple in the future).
+type Unit struct {
+	id       string
+	unitType UnitType
+
+	expLock   sync.RWMutex
+	exp       UnitState
+	config    string
```
ph, line 53: This would be the string representation of the yaml, this seems asking for having his own type? We we can validate the presence of some fields depending on the unit or that we have valid yaml.

ph: Sadly we can have a `yaml.RawMessage` because of the way indent works and could break the configuration.

ph (changes requested):
I've added a few questions in the PR, and so far it looks good to me.
I've been having a few difficulties concerning the Keystore API and implementation, but I find it hard without playing in the context of the agent.

Looking at other comments concerning renaming `txn` to `tx` I believe is a good thing to do.

elastic-agent-client.proto
blakerouse: @aleksmaus the state of the connection will be tracked on the server-side (aka. Elastic Agent), when the connection is lost then the transaction will be discarded server-side. At that point that transaction is broken and they client will need to handle the error and create a new transaction

I also think Elastic Agent should place a hard timeout on transactions, as it would not be good to keep them open for a long period of time. They really should be short in time.

I will go through and adjust it all from "Txn" abbr. to "Tx".

pkg/client/client_v2.go
blakerouse: After looking at the golang I do not believe this to be an issue. The ticker will keep ticking, on hibernate it will just pickup where it left off because it uses the monoclock. The ticker also handles a slow reader so it never adds more than one to the channel so until the next is removed from the channel it will not add another one.

pkg/client/client_v2_test.go
blakerouse: For testing make it constant? For the library, no I would prefer they provide more detail than that. Like "Connected to elasticsearch(https://localhost:8200)" or something.

pkg/client/common.go
blakerouse: Not an issue with the usage of ticker.

pkg/client/common.go
blakerouse: Will do.

pkg/client/unit.go
blakerouse: Is it YAML, JSON, or something else. The point of the client is that it doesn't know and doesn't care. So its best to keep it just as a string representation that could then be decoded.

pkg/client/client_v2_test.go
blakerouse: Its in a seperate goroutine, `go vet` will not allow this.

`call to (*T).Fatal from a non-test goroutine`

pkg/client/client_v2_test.go
ph: OK good point.

ph (approved):
LGTM, Let's get another approval from @elastic/elastic-agent-data-plane

elastic-agent-client.proto
```diff
@@ -25,6 +41,8 @@ message ConnInfo {
     bytes peer_cert = 5;
     // Peer private key.
     bytes peer_key = 6;
+    // Allowed services that spawned process can use. (only used in V2)
+    repeated ConnInfoServices services = 7;
```
cmacknz, line 45: What is it expected that processes do with the services list? Is the intent that each process checks to make sure the services it needs are actually available?

pkg/client/client_v2.go
```diff
@@ -0,0 +1,543 @@
...
+}
+
+// UnitChanges returns channel client send unit change notifications to.
+func (c *clientV2) UnitChanges() <-chan UnitChanged {
+	return c.unitsCh
+}
+
+// Errors returns channel of errors that occurred during communication.
+func (c *clientV2) Errors() <-chan error {
```
cmacknz, line 143: Can you document that failing to read `Errors()` or `UnitChanges()` will cause the client to block internally?

cmacknz (approved):
LGTM, I expect we will have more feedback once we actually try to use the new V2 client :)

ph:
@cmacknz I also change when we try the real world.

fearful-symmetry:
Agreed on real-world testing, I'm honestly kinda planning with the assumption that a lot of the APIs are gonna change over time as we actually use it.

elastic-agent-client.proto
blakerouse: Correct. A process can only use the services that it is given, if it tries to Elastic Agent will return an error back to the caller.

pkg/client/client_v2.go
blakerouse: It is documented on the interface.

PR #34, opened 2022-07-28
[v2] Add expected config and agent info to v2 protocol.

blakerouse:
Adds the `AgentInfo` to the first check-in expected message to the client. Adds `AgentInfo` function to the `ClientV2` interface to allow the user of the client to get the `AgentInfo` for the agent that the client is connected to.

Adds `UnitLogLevel` to `UnitExpected`. The log level of the unit is now a top-level item for a unit. This will allow the Elastic Agent to control the log-level of the unit both through configuration or by toggling the log-level though its own daemon control protocol.

Changes `config string` to `config UnitExpectedConfig` on `UnitExpected`. This change ensures that the client user never has to decode yaml to understand the configuration. `source` fields exists on all messages to ensure that all data is always present. Some fields are defined and are top-level (but the source for that is always present in `source` for the message) for convenience.


elastic-agent-client.proto
```diff
@@ -177,6 +178,78 @@ enum UnitType {
     OUTPUT = 1;
 }
 
+// Log level for the unit.
+enum UnitLogLevel {
+    TRACE = 0;
```
cmacknz, line 183: This will be the default trace level if this field is left unset, should you reverse the order to be safe?

cmacknz: It already doesn't exactly match the levels in beats https://pkg.go.dev/github.com/elastic/elastic-agent-libs@v0.2.9/logp#Level

elastic-agent-client.proto
```diff
@@ -177,6 +178,78 @@ enum UnitType {
...
+    string namespace = 4;
+}
+
+// Stream defined in a configuration.
+message Stream {
+    // Source is the original source of the message. All values from the message are included here
+    // even if other concrete fields are defined for this message.
+    google.protobuf.Struct source = 1;
+    string id = 2;
+    DataStream data_stream = 3;
```
cmacknz, line 229: Is it worth including a repeated list of processors? Likely the type would have to be `google.protobuf.Struct` as I don't think we want to encode the beats processor structure here, so it probably just saves us a key looking in `source`.

elastic-agent-client.proto
```diff
@@ -189,12 +262,29 @@ message UnitExpected {
...
+}
+
+// Only provided on first checkin expected response to the component.
+//
+// Includes the agent information that the component might want to use for its events.
+message CheckinAgentInfo {
+    // ID is the Elastic Agent's unique ID.
+    string id = 1;
+    // Version is the version of the running Elastic Agent.
+    string version = 2;
```
cmacknz, line 277: Is it worth having this be a struct with major.minor.patch fields?

elastic-agent-client.proto
blakerouse: If we want to define a proper message for `processors` then I would be okay with placing it in here. I believe the policy should be its either a proper message or you look for it in `source`. There is no real benefit unless its a proper message.

elastic-agent-client.proto
blakerouse: Both Elastic Agent and Fleet Server just use version as a string. I think doing something like that is over the top.

elastic-agent-client.proto
blakerouse: I was basing it off of zerolog but off by 1. Allowing it to be the `log_level - 1` to get the log level for zerolog. Why zerolog? Because zerolog is what we use in `fleet-server` and what we want to transition the Elastic Agent to using as well.

I think the numbers don't really matter. Would be happy to change it, maybe even place 0 as error. So that the default if not present is only error.

elastic-agent-client.proto
fearful-symmetry: Yah, the processors are inconsistent enough that we'd have to define all of them, which would be an...annoying process. I'm leaning towards just not defining them, since I don't think a `[]string` is super useful. We can add it later if needed, though.

ph: I agree with @fearful-symmetry there are indeed all different here, I don't think we need it initially so I would be more relax for it.

elastic-agent-client.proto
ph: I would keep it as a string, IIRC we a type that could encapsulate the version and allow us to do any kind of comparison.

ph (approved):
This look ok to me, anything @cmacknz @fearful-symmetry that must be here?

pkg/client/unit.go
```diff
@@ -76,10 +93,10 @@ func (u *Unit) Type() UnitType {
 }
 
 // Expected returns the expected state and config for the unit.
-func (u *Unit) Expected() (UnitState, string) {
+func (u *Unit) Expected() (UnitState, UnitLogLevel, *proto.UnitExpectedConfig) {
```
fearful-symmetry, line 96: Is there a reason why the log level is being sent as a separate return value?

elastic-agent-client.proto
cmacknz: I like having the default of 0 map to error, to avoid accidental log spam in case this is ever omitted.

fearful-symmetry:
Have one clarifying comment, other than that looks good. @blakerouse I assume the next step after this is making the change in the elastic-agent side?

cmacknz (approved):

pkg/client/unit.go
blakerouse: This is moved to top-level because it will allow the Elastic Agent to change a units log level without modifying the configuration that comes from the policy.

The idea here is that with Elastic Agent running in any mode, you can use the Elastic Agent CLI to call into the running daemon on change a unit to say trace. That will allow debug-ability of that single unit, without having to modify the policy.

I still think that Elastic Agent can read `log_level` from the configuration as the default `log_level` for a unit, but that will come here in a top-level and not shown in the configuration, just like `use_output`. Then for debugging purposes that `log_level` can be changed different from what is in the policy.

elastic-agent-client.proto
blakerouse: I agree. I am flipping this enum to make error 0.

PR #35, opened 2022-08-04
[v2] Change unit observed payload to google.protobuf.Struct.

blakerouse:
With the changed in #34 to make the configuration a defined type and no longer a YAML encoded string. This makes a similar change for `payload` on `UnitObserved`.

It was previously a JSON encoded string, but with the transition to using `google.protobuf.Struct` inside of the `UnitExpectedConfig` there is no reason not to do the same for `payload`.


elastic-agent-client.proto
```diff
@@ -306,7 +306,7 @@ message UnitObserved {
     // Exposed to users to provide more detail about the state for this single unit.
     string message = 5;
     // JSON encoded payload for the state.
```
michel-laterman, line 308: change the comment as well?

michel-laterman (approved):

elastic-agent-client.proto
blakerouse: Yes, will do!

PR #39, opened 2022-09-07
Add ability to call diagnostics with V2 protocol and add defaults.

blakerouse:
## Overview

Completes the TODO for supporting the new diagnostics for the V2 protocol. With this change any component that uses the client gets all the defaults from the runtime/pprof included.

The client can have diagnostics that each unit will inherit and then each unit can define its own diagnostic as well.

Elastic Agent will call this per unit to provide all the diagnostics information for the running components unit.

### DIFF
--- a/pkg/client/client_v2.go
+++ b/pkg/client/client_v2.go
@@ -10,13 +10,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"google.golang.org/protobuf/types/known/timestamppb"
 	"io"
 	"runtime/pprof"
 	"sync"
 	"time"
 
 	"google.golang.org/grpc"
+	"google.golang.org/protobuf/types/known/timestamppb"
 
 	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
 	"github.com/elastic/elastic-agent-client/v7/pkg/utils"
--- a/pkg/client/client_v2.go
+++ b/pkg/client/client_v2.go
@@ -5,10 +5,14 @@
 package client
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
+	"google.golang.org/protobuf/types/known/timestamppb"
 	"io"
+	"runtime/pprof"
 	"sync"
 	"time"
 
@@ -89,6 +93,10 @@ type V2 interface {
 	//
 	// nil can be returned when the client has never connected.
 	AgentInfo() *AgentInfo
+	// RegisterDiagnosticHook registers a diagnostic hook function that will get called when diagnostics is called for
+	// as specific unit. Registering the hook at the client level means it will be called for every unit that has
+	// diagnostics requested.
+	RegisterDiagnosticHook(name string, description string, filename string, contentType string, hook DiagnosticHook)
 }
 
 // clientV2 manages the state and communication to the Elastic Agent over the V2 control protocol.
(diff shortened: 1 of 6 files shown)


pkg/client/client_v2.go
```diff
@@ -89,6 +93,10 @@ type V2 interface {
 	//
 	// nil can be returned when the client has never connected.
 	AgentInfo() *AgentInfo
+	// RegisterDiagnosticHook registers a diagnostic hook function that will get called when diagnostics is called for
+	// as specific unit. Registering the hook at the client level means it will be called for every unit that has
```
cmacknz, line 97:
```suggestion
	// a specific unit. Registering the hook at the client level means it will be called for every unit that has
```

pkg/client/client_v2.go
```diff
@@ -581,6 +602,80 @@ func (c *clientV2) tryPerformAction(actionResults chan *proto.ActionResponse, ac
...
+	diagHooks := make(map[string]diagHook)
+	c.dmx.RLock()
+	for n, d := range c.diagHooks {
+		diagHooks[n] = d
+	}
+	c.dmx.RUnlock()
+
+	// unit hooks after client hooks; allows unit specific to override client hooks
+	unit.dmx.RLock()
+	for n, d := range unit.diagHooks {
```
cmacknz, line 630: Is there a reason not to do this in the same lock protected block as `for n, d := range c.diagHooks` above? I assume it is intentionally the same mutex for both maps?

It's not incorrect I just wanted to double check the intent here.

pkg/client/client_v2.go
```diff
@@ -581,6 +602,80 @@ func (c *clientV2) tryPerformAction(actionResults chan *proto.ActionResponse, ac
...
+		diagHooks[n] = d
+	}
+	unit.dmx.RUnlock()
+
+	// perform diagnostics in goroutine, so we don't block other work
+	go func() {
+		res := make([]*proto.ActionDiagnosticUnitResult, 0, len(diagHooks))
+		for n, d := range diagHooks {
+			content := d.hook()
```
cmacknz, line 639: Is it worth considered hooks that could block or time out? Or do we expect them to always complete quickly? Is it worth passing a done channel or a context through to the hooks?

cmacknz (commented):
A few minor comments, mostly LGTM.

We will need to have everyone update their agent client version for this to work, we'll need follow up TODOs for this.

pkg/client/client_v2.go
blakerouse: Different lock is used because each unit can have its own hooks and can be registered at any time. So we need to have a unique lock per unit for hook registration.

pkg/client/client_v2.go
blakerouse: I think we should leave that up to the hook. I think encoding the timeout at this level would mean you have to register with a timeout, but if a hook needs to manage that timeout it could just do that in its own function.

I would hope that most are really fast and can quickly get the results back to the Elastic Agent. The Elastic Agent will place a timeout on the Diagnostic request to the component so if it takes longer than that timeout to return then the Elastic Agent will just continue on.

cmacknz: Thanks I just misread this, I was focused on `dmx.RLock()` and missed that it was `unit.dmx.RLock()` on the second loop

cmacknz (approved):
LGTM, leave this up for a day or so to give the others time to review. They may find something I can't see.

Issue #44, opened 2022-10-14
Should we differntiate between V2 Unit states?


fearful-symmetry:
So, a footcannon that I've shot myself with a few times is mixing up what Unit states will be sent to the client via the server, and what unit states are purely for a client to report status:

https://github.com/elastic/elastic-agent-client/blob/3e014f7c0e838727f2596e3463f5110590e529cf/pkg/proto/elastic-agent-client.pb.go#L94-L102

For example, `State_STOPPING` will only ever be sent by a client to the server to update its status, but there's nothing in the code to suggest this, and so it's easy to treat it like a state that will be requested by `Expected`, leading to coding errors.

I wonder if the code in `unit.go` should break out the states into multiple types, for example a `UnitStateExpected` and a `UnitStateReported` just do make this distinction clear at a type level. Cc @cmacknz

cmacknz:
I tried to address this with documentation of the unit states https://github.com/elastic/elastic-agent-client/pull/45

I agree this needs to be clarified, I've seen this confusion in the unit/beat PRs as well.

cmacknz:
Closing this, this is now explicitly documented in the proto file: https://github.com/elastic/elastic-agent-client/pull/45

PR #45, opened 2022-10-17
Improve the description of the possible unit states.

cmacknz:
Clarify which states are optional and intended as observed states from a unit, and which states will be sent by the agent as expected states. Unit authors have been confused about which states can be received in `UnitExpected`. 

Alternatively we could use separate state enumerations for the expected and observed messages like the v1 protocol does: https://github.com/elastic/elastic-agent-client/blob/3e014f7c0e838727f2596e3463f5110590e529cf/elastic-agent-client.proto#L84-L92

Follow up from this PR comment https://github.com/elastic/beats/pull/32673#discussion_r959865366

- Closes # https://github.com/elastic/elastic-agent-client/issues/44


fearful-symmetry:
> Alternatively we could use separate state enumerations for the expected and observed messages like the v1 protocol does

Yah, I'm more a fan of this idea, if we can enforce this at the type level, it'll be harder to typo things like `STOPPING` vs `STOPPED`.

blakerouse (approved):
Looks good!

PR #47, opened 2022-11-18
Change struct compare statement in unit update

fearful-symmetry:
First PR to fix https://github.com/elastic/elastic-agent/issues/1738 and https://github.com/elastic/apm-server/pull/9544#issuecomment-1316473137

This attempts to fix an underlying issue where V2 clients are seeing unit update events where the config has apparently remained unchanged. In order for that change event to be generated, The underlying unit needs to have an incremented config index, and a different config. I'm not sure why we're seeing an incremented index, but this at least fixes how we try to compare the structs. Instead of `!=` on two pointers, this does a `reflect.DeepEqual` on the underlying source config.

### DIFF
--- a/pkg/client/unit.go
+++ b/pkg/client/unit.go
@@ -8,10 +8,10 @@ import (
 	"reflect"
 	"sync"
 
+	gproto "google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
-	"github.com/elastic/elastic-agent-libs/logp"
 )
 
 // UnitType is the type of the unit, either input or output
@@ -252,7 +252,6 @@ func (u *Unit) RegisterDiagnosticHook(name string, description string, filename 
 
 // updateConfig updates the configuration for this unit, triggering the delegate function if set.
 func (u *Unit) updateState(exp UnitState, logLevel UnitLogLevel, cfg *proto.UnitExpectedConfig, cfgIdx uint64) bool {
-	log := logp.L()
 	u.expMu.Lock()
 	defer u.expMu.Unlock()
 	changed := false
@@ -266,7 +265,7 @@ func (u *Unit) updateState(exp UnitState, logLevel UnitLogLevel, cfg *proto.Unit
 	}
 	if u.configIdx != cfgIdx {
 		u.configIdx = cfgIdx
-		if !reflect.DeepEqual(u.config.Source.AsMap(), cfg.Source.AsMap()) {
+		if !gproto.Equal(u.config.Source.ProtoReflect().Interface(), cfg.Source.ProtoReflect().Interface()) {
 			u.config = cfg
 			changed = true
 		}
--- a/pkg/client/unit.go
+++ b/pkg/client/unit.go
@@ -265,7 +265,7 @@ func (u *Unit) updateState(exp UnitState, logLevel UnitLogLevel, cfg *proto.Unit
 	}
 	if u.configIdx != cfgIdx {
 		u.configIdx = cfgIdx
-		if !gproto.Equal(u.config.Source.ProtoReflect().Interface(), cfg.Source.ProtoReflect().Interface()) {
+		if !gproto.Equal(u.config.Source, cfg.Source) {
 			u.config = cfg
 			changed = true
 		}
--- a/pkg/client/unit.go
+++ b/pkg/client/unit.go
@@ -11,6 +11,7 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 
 	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
+	"github.com/elastic/elastic-agent-libs/logp"
 )
 
 // UnitType is the type of the unit, either input or output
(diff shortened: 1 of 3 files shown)


pkg/client/unit.go
```diff
@@ -264,7 +266,7 @@ func (u *Unit) updateState(exp UnitState, logLevel UnitLogLevel, cfg *proto.Unit
 	}
 	if u.configIdx != cfgIdx {
 		u.configIdx = cfgIdx
-		if u.config != cfg {
+		if !reflect.DeepEqual(u.config.Source.AsMap(), cfg.Source.AsMap()) {
```
cmacknz, line 269: Since `config.Source` is a a proto message of type `google.protobuf.Struct` you should compare them using https://pkg.go.dev/google.golang.org/protobuf/proto#Equal

Comparing as maps sort of works around this but I am not sure how reliable it is. Each instance of a protobuf message contains several internal fields (serialization caches, lists of unparseable fields, etc) that aren't necessarily stable that the comparison needs to know to ignore

cmacknz (commented):
Is there a test you can write to reproduce this and prove that the fix works as expected?

pkg/client/unit.go
cmacknz: This is probably causing the unexpected index updates. Not being able to compare proto messages in the obvious ways is one of the primary annoyances of proto messages in Go unfortunately.

Usually the best way to detect these problems is to do a comparison, then marshal one of the messages to populate the `sizeCache`, then compare the messages again and it will fail unless you used `proto.Equal`.

fearful-symmetry: > Comparing as maps sort of works around this but I am not sure how reliable it is. Each instance of a protobuf message contains several internal fields (serialization caches, lists of unparseable fields, etc) that aren't necessarily stable that the comparison needs to know to ignore

Yah, that's why I did `Souce.AsMap()`, since that'll just return the public, exported fields, and not any of the internal protobuf struct fields.  Didn't know the protobuf library has its own `Equal` function, should probably switch to that though.

fearful-symmetry: Yah, agreed that `poto.Equal` is definitely a more sound choice here.

And yah, after I update this PR I can open a similar once against elastic-agent.

cmacknz: You also need the client version bump here updated in Beats, Fleet Server, and possibly APM server in addition to Agent.

fearful-symmetry:
Alright, updated to use `proto.Equal`, added some paranoid tests.

pkg/client/unit.go
```diff
@@ -264,7 +265,7 @@ func (u *Unit) updateState(exp UnitState, logLevel UnitLogLevel, cfg *proto.Unit
 	}
 	if u.configIdx != cfgIdx {
 		u.configIdx = cfgIdx
-		if u.config != cfg {
+		if !gproto.Equal(u.config.Source.ProtoReflect().Interface(), cfg.Source.ProtoReflect().Interface()) {
```
cmacknz, line 268:
```suggestion
		if !gproto.Equal(u.config.Source, cfg.Source) {
```

Shouldn't this work? Why is the additional `.ProtoReflect().Interface()` call needed? I just did a quick test and this builds, but I didn't run it though the tests.

cmacknz: Yep tests still pass with this change.

fearful-symmetry: Ah, my bad, was poking around the `proto` library a bit, forgot to revert that.

pkg/client/unit_test.go
```diff
@@ -0,0 +1,58 @@
...
+	sameMap := map[string]interface{}{"username": "test"}
+	pbStruct, err := structpb.NewStruct(sameMap)
+	require.NoError(t, err)
+	defaultTest.config.Source = pbStruct
+
+	pbStructNew, err := structpb.NewStruct(sameMap)
+	newUnit := &proto.UnitExpectedConfig{
+		Source: pbStructNew,
+	}
+	// This should return false, as the two underlying maps in `source` are the same
```
cmacknz, line 29: If you add a call to `proto.Marshal` it would fail if you fell back to using `reflect.DeepEqual`. I think it's worth adding this to ensure the implementation doesn't change later.

```diff
diff --git a/pkg/client/unit_test.go b/pkg/client/unit_test.go
index e4539a6..610cfc5 100644
--- a/pkg/client/unit_test.go
+++ b/pkg/client/unit_test.go
@@ -5,6 +5,7 @@ import (

        "github.com/elastic/elastic-agent-client/v7/pkg/proto"
        "github.com/stretchr/testify/require"
+       gproto "google.golang.org/protobuf/proto"
        "google.golang.org/protobuf/types/known/structpb"
 )

@@ -26,6 +27,11 @@ func TestUnitUpdateWithSameMap(t *testing.T) {
        newUnit := &proto.UnitExpectedConfig{
                Source: pbStructNew,
        }
+
+       // Marshal the message to populate the size cache and ensure the internal proto files are populated.
+       _, err = gproto.Marshal(newUnit)
+       require.NoError(t, err)
+
        // This should return false, as the two underlying maps in `source` are the same
        result := defaultTest.updateState(UnitStateHealthy, UnitLogLevelDebug, newUnit, 2)
        require.False(t, result)
(END)
```

cmacknz: (I confirmed this test does fail if you switch the comparison to use `reflect.DeepEqual` instead of `proto.Equal`)

fearful-symmetry: Ahh, didn't even think of that.

cmacknz: I have had this bug before :)

cmacknz (approved):

PR #48, opened 2022-11-24
Remove elastic/elastic-agent dependency

axw:
The main go.mod requires elastic/elastic-agent because it is used by v2tool. This causes users of elastic-agent-client to fetch the elastic-agent repo unnecessarily. This is doubly bad because the version of elastic-agent required does not exist; it probably came from a branch? For example, this is causing errors in APM Server's CI:

https://apm-ci.elastic.co/job/apm-server/job/apm-server-mbp/job/PR-9667/3/consoleFull

> 15:09:54  go: github.com/elastic/elastic-agent@v0.0.0-20220831162706-5f1e54f40d3e: invalid version: unknown revision 5f1e54f40d3e

Move v2tool to its own module to prevent this dependency leaking to users of elastic-agent-client. Update the elastic-agent module version, and adjust the v2tool code accordingly.


cmacknz (approved):
Thanks! v2tool probably needs to move into the elastic-agent repository, I think we may have a branch for this that isn't complete yet.

efd6:
The root cause that this fixes for elastic-agent-client breaks the beats build. The relevant change for this repo that brings in the out-of-branch elastic-agent commit is https://github.com/elastic/elastic-agent-client/commit/3e014f7c0e838727f2596e3463f5110590e529cf.

PR #50, opened 2022-12-13
Add a filemode to the v2 client

fearful-symmetry:
This PR is the final part of my ON week project to improve overall debugability in the agent client ecosystem. This does a few things: 

- Add a standalone filemode to the elastic-agent-client, a client can call `NewV2FromHardcodedConfig` with a filepath, the client will read this file and use it to generate a set of units sent to the client. This allows the client to operate in standalone mode, without elastic agent. This makes quick debugging and development much easier, as there's no need to run agent alongside whatever client you're working on. I've already used this to debug a few issues in beats under agent. This mode has a few limits, as most other features (artifact store, etc) are disabled.
- This migrates the `ExpectedConfig` set of functions from elastic-agent's internal libs to here. I think this is a good fit for this set of functions, and its needed for the filemode.
- Add a few logging lines for unit change events. These log lines tend to get yak-shaved into any V2 clients, so moving them here improves ergonomics. We still have a long way to go as far as debugging unit change behavior, though.

All filemode behavior is disabled by default and must be explicitly implemented by any clients, so this should produce no changes (aside from log lines) in existing clients.


cmacknz:
Closing, not needed for 5 months :)

PR #51, opened 2022-12-20
changed namespace to be unique

michalpristas:
The problem is that in order to fix upgrade scenario in agent upgrading from V1 to V2 agent we need to provide backward compatibility

Both v1 control and this elastic agent client contains multiple shared field which were added as part of V2 work.
These field are e.g enum STOPPING, STARTING... which needs to be unique inside a package.

Both v1 control and this elastic agent client use same namespace, we changed that to cproto in v2 control protocol but for backward compatibility I cannot change v1 control protocol. 

Having namespace !- proto seems also like a sane thing, and as components communicate with the agent in same version always (as for now) it should not even be breaking. 

Note: all components need to be updated prior to 8.6


elastic-agent-client.proto
```diff
@@ -4,7 +4,7 @@
 
 syntax = "proto3";
 
-package proto;
+package agentclient;
```
cmacknz, line 7: You can put a version in this I believe, like `agent.client.v1` and `agent.client.v2`

cmacknz: Maybe `agent.control.v1` and v2

cmacknz: We shouldn't use client really since this will generate both a client and server

PR #52, opened 2022-12-21
Always send observed after removals

blakerouse:
# Overview

This changes the logic to always to send the observed state once a unit has been removed.

# Why is this needed?

With out this change when units are removed the observed state is not sent to the Elastic Agent until the next checkin timeout which can be up to 25 seconds.


cmacknz (approved):
There should probably be a test for this ;)

blakerouse:
@cmacknz Its rather hard to test and the reason the tests didn't catch this is because the `minCheckTimeout` is adjusted in the tests to make it much faster to test.

PR #54, opened 2023-01-24
Use GetSource() calls to compare structs in updateState()

fearful-symmetry:
Fixes a panic I noticed while writing tests, we can get a nil reference if we forget to include a source field. This just uses the proper `Get*()` calls so we can avoid that.


cmacknz (approved):
The source field should always be present so I'm curious as to how that was happening but the change is a good idea regardless.

PR #56, opened 2023-02-01
Add feature flag

### LINKED ISSUE #55: Add feature flags to the control protocol
Modify the [control protocol](https://github.com/elastic/elastic-agent-client/blob/main/elastic-agent-client.proto) and the [client](https://github.com/elastic/elastic-agent-client/blob/main/pkg/client/client_v2.go) to handle feature flags defined on https://github.com/elastic/elastic-agent/issues/2185

AndersonQ:
closes #55 
relates https://github.com/elastic/elastic-agent/issues/2185

## Testing notes

See testing notes on https://github.com/elastic/elastic-agent/pull/2218.


elastic-agent-client.proto
```diff
@@ -261,6 +261,8 @@ message UnitExpectedConfig {
     DataStream data_stream = 7;
     // Multiple streams per unit.
     repeated Stream streams = 8;
+    // Features configure, enable and disable optional features.
+    google.protobuf.Struct features = 9;
```
cmacknz, line 265: Why `google.protobuf.Struct`? This is going to force every client to parse the feature as JSON effectively. Can we use a custom type with more structure?

I see some value in using struct because if we do not necessarily want to have to recompile the agent every time a feature gets added. It is maybe worth asking if that makes sense though, as features that only apply to particular integrations can just go in the integrations config. Can we think of features that would apply to multiple processes run by agent where the agent itself would require no changes?

If we wanted to allow the addition of features without recompiling the agent, we could still do something more structured like the following:

```protobuf
// Map instead of a repeated message so we can give each feature a name
map<string, Feature> features = 9;
```

```protobuf
message Feature {
  google.protobuf.Struct source = 1;
  // Pull out the enabled flag for convenience since it is always necessary.
  bool enabled = 1;
}

If we didn't want the ability to pass through features such that we expect to always have to change the agent, then something like this would work:

```protobuf
Features features = 9;
```

```protobuf
message Features {
  FQDNFeature fqdn = 1;
}

message FQDNFeature {
  bool enabled = 1;
}
```

blakerouse: I question if feature flags should be per units.

I think it might be better to make this a top level key with a top level message type.

```
// A set of units and their expected states and configuration.
message CheckinExpected {
    // Units is the expected units the component should be running.
    repeated UnitExpected units = 1;
    // Agent info is provided only on first CheckinExpected response to the component.
    CheckinAgentInfo agent_info = 2;

    Features features = 3;
}
```

blakerouse (changes requested):

pkg/client/client_v2.go
```diff
@@ -356,6 +356,7 @@ func (c *clientV2) sendObserved(client proto.ElasticAgent_CheckinV2Client) error
 
 // syncUnits syncs the expected units with the current state.
 func (c *clientV2) syncUnits(expected *proto.CheckinExpected) {
+	// here!!! send happens here
```
michalpristas, line 359: ???

AndersonQ: debug comments

pkg/client/unit.go
```diff
@@ -146,11 +147,11 @@ func (u *Unit) Type() UnitType {
 	return u.unitType
 }
 
-// Expected returns the expected state and config for the unit.
-func (u *Unit) Expected() (UnitState, UnitLogLevel, *proto.UnitExpectedConfig) {
+// Expected returns the expected state, log leve, features and config for the unit.
+func (u *Unit) Expected() (UnitState, UnitLogLevel, *proto.Features, *proto.UnitExpectedConfig) {
```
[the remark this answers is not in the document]
cmacknz: This is just the Go client, only changes in the .proto file affect endpoint. That said this feature will not fully work unless we align with endpoint, cloudbeat, and APM as well.

elastic-agent-client.proto
```diff
@@ -292,12 +292,22 @@ message CheckinAgentInfo {
...
+}
+
 // A set of units and their expected states and configuration.
 message CheckinExpected {
     // Units is the expected units the component should be running.
     repeated UnitExpected units = 1;
     // Agent info is provided only on first CheckinExpected response to the component.
     CheckinAgentInfo agent_info = 2;
+
+    Features features = 3;
```
cmacknz, line 310: Add comments explaining what this is, what the format is, and then document the FQDN feature itself and what clients should do with it please :)

pkg/client/unit.go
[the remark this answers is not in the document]
cmacknz: I'm also not sure the features makes sense as a per unit return value here, since the feature flags are not actually per unit. 

A separate `Features()` channel seems like a better approach, that way users of this don't need to filter out and separately handle the feature changes from the unit changes.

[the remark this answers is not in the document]
michalpristas: that's what we were talking with anderson in the mornign. providing a separate channel to watch inside checkin loop

pkg/client/unit.go
```diff
@@ -251,25 +252,37 @@ func (u *Unit) RegisterDiagnosticHook(name string, description string, filename
...
 	if u.logLevel != logLevel {
 		u.logLevel = logLevel
 		changed = true
 	}
+
 	if u.configIdx != cfgIdx {
 		u.configIdx = cfgIdx
-		if !gproto.Equal(u.config.GetSource(), cfg.GetSource()) {
+		if !gproto.Equal(u.config.GetSource(), cfg.GetSource()) ||
+			!gproto.Equal(u.features, features) {
```
cmacknz, line 278: I think this will be a source of bugs, because we will now send unchanged unit configurations to clients of this code and they will need to filter them out.

The intent of the elastic agent client is to only send unit modified events to a client of the package when the actual unit configuration changed. This breaks that contract, and will send unchanged unit configurations if only the feature flags have been modified.

cmacknz (changes requested):
I don't think the feature flags have been handled in the right place in the client code.

pkg/client/unit.go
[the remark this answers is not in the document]
cmacknz: Yep, what is done now I think breaks the client API contract https://github.com/elastic/elastic-agent-client/pull/56#discussion_r1098692116

We need to change this before we can merge it.

brian-mckinney (approved):
My review applies only to the .proto changes.  Adding the Features as a top level field will work better for Endpoint as well.  👍

cmacknz:
The team spoke about what we should do in the Go client this morning and concluded:

1. When a feature changes, we should trigger a UnitModified event for every unit.
2. The UnitModified event should be extended to included the set of changes that triggered the event. The set of possible change triggers right now is a change to a Feature, a change to the log level, and a change to the Unit configuration. A single UnitModified event must be associated with at least one of these events, but may also include all of them together if all of them have changed.

We agreed that the changes to the .proto files are good as they are.

pkg/client/unit.go
```diff
@@ -181,7 +181,9 @@ func (u *Unit) UpdateState(state UnitState, message string, payload map[string]i
 		u.stateMsg = message
 		changed = true
 	}
-	if (u.statePayload == nil && statePayload != nil) || (u.statePayload != nil && statePayload == nil) || !reflect.DeepEqual(u.statePayload, statePayload) {
+	if (u.statePayload == nil && statePayload != nil) ||
```
cmacknz, line 184: nit: the reformatting of code you didn't actually change is making this review harder to do :)

pkg/client/client_v2.go
```diff
@@ -365,42 +393,90 @@ func (c *clientV2) syncUnits(expected *proto.CheckinExpected) {
...
+				agentUnit.Config,
+				agentUnit.ConfigStateIdx)
+			if triggers != nil { // a.k.a something changed
+				changed := UnitChanged{
+					Triggers: triggers,
+					Type:     UnitChangedModified,
+					Unit:     unit,
 				}
+
+				changed = c.syncFeatures(expected, changed)
```
cmacknz, line 446: This should be updating the triggers list shouldn't it?

Also since Features exist at the top level of the protocol and is not actually per unit, it feels like this should happen outside the `for _, agentUnit := range expected.Units {` loop above and then the result and trigger (if needed) copied into each unit.

Otherwise I suspect you'd only detect a feature change for the first unit in this list, and omit the trigger from the others unless you kept a copy of the Features for each unit individually.

pkg/client/client_v2.go
```diff
@@ -79,15 +100,15 @@ type V2 interface {
 	Start(ctx context.Context) error
 	// Stop stops the connection to Elastic Agent.
 	Stop()
-	// UnitChanges returns channel client send unit change notifications to.
+	// UnitChanged returns the channel the client send change notifications to.
 	//
 	// User of this client must read from this channel, or it will block the client.
-	UnitChanges() <-chan UnitChanged
+	UnitChanged() <-chan UnitChanged
```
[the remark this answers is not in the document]
cmacknz: You will need to follow through with this change in fleet-server, elastic-agent-shipper, and beats at a minimum. There are some other inputs being developed in Go by security and observability teams that would also need to handle this the next time they update.

pkg/client/client_v2_test.go
```diff
@@ -125,12 +129,19 @@ func TestClientV2_Checkin_Initial(t *testing.T) {
...
 	go func() {
 		for {
 			select {
 			case <-ctx.Done():
 				return
-			case change := <-validClient.UnitChanges():
+			case change := <-validClient.UnitChanged():
+				for _, t := range change.Triggers {
+					if t == TriggerFeature {
+						gotFQDN = change.Features.FQDN.Enabled
```
cmacknz, line 141: Is it worth having a trigger per feature? 

It seems like as soon as we support more than one feature we will force every user of this package to have to iterate over all of them to find what changed. Having something like `TriggerFQDNFeature` instead seems like it would be more convenient, at the cost of an additional enum entry per feature.

pkg/client/client_v2_test.go
```diff
@@ -284,12 +301,20 @@ func TestClientV2_Checkin_UnitState(t *testing.T) {
...
+	var gotFQDN bool
+	var gotTriggers []Trigger
 	go func() {
 		for {
 			select {
 			case <-ctx.Done():
 				return
-			case change := <-client.UnitChanges():
+			case change := <-client.UnitChanged():
+				for _, t := range change.Triggers {
```
cmacknz, line 312: We should be testing that each of the trigger types works, not just the feature triggers.

pkg/client/unit.go
```diff
@@ -250,27 +252,37 @@ func (u *Unit) RegisterDiagnosticHook(name string, description string, filename
...
 		u.configIdx = cfgIdx
 		if !gproto.Equal(u.config.GetSource(), cfg.GetSource()) {
 			u.config = cfg
-			changed = true
+			triggers = append(triggers, TriggerConfig)
 		}
 	}
-	return changed
+
+	return triggers
```
cmacknz, line 285: We need a unit test of the trigger detection logic.

cmacknz (commented):
API design is looking good, minus a few nits and missing tests.

pkg/client/client_v2.go
```diff
@@ -22,35 +22,56 @@ import (
...
 }
 
-// UnitChanged is what is sent over the UnitChanges channel any time a unit is added, modified, or removed.
+// UnitChanged is what is sent over the UnitChanged channel any time a change happens:
+//   - a unit is added, modified, or removed
+//   - a feature changes
 type UnitChanged struct {
-	Type UnitChangedType
+	Type     ChangeType
+	Triggers []Trigger
```
blakerouse, line 70: I would make this into a bitmask, like we discussed in the call.

elastic-agent-client.proto
```diff
@@ -292,12 +292,28 @@ message CheckinAgentInfo {
...
 // A set of units and their expected states and configuration.
 message CheckinExpected {
     // Units is the expected units the component should be running.
     repeated UnitExpected units = 1;
     // Agent info is provided only on first CheckinExpected response to the component.
     CheckinAgentInfo agent_info = 2;
+
+    // Features are the expected feature flags configurations.
+    // Added on Elastic Agent v8.7.1.
+    Features features = 3;
```
blakerouse, line 316: Lets add the `unit64 features_idx = 4;`to this expected states. Also add the `features_idx` to the observed state.

pkg/client/artifact_test.go
```diff
@@ -65,7 +65,7 @@ func TestArtifact(t *testing.T) {
 			select {
 			case <-ctx.Done():
 				return
-			case change := <-client.UnitChanges():
+			case change := <-client.UnitChanged():
```
blakerouse, line 68: Lets not change the name of the method.

pkg/client/client_v2.go
```diff
@@ -22,35 +22,56 @@ import (
...
+// FeatureFQDN indicates if the FQDN should be used instead of hostname for host.name.
+// See proto.FQDNFeature for details.
+type FeatureFQDN struct {
+	Enabled bool
+}
+
+// Features are configurations for all the features (flag) available.
+// See proto.Features for details.
+type Features struct {
+	FQDN FeatureFQDN
```
blakerouse, line 62: Lets not added any defined structs for features in the client. That should all come from the protocol, don't want to have to modify the client code to add a new feature.

elastic-agent-client.proto
```diff
@@ -292,12 +292,28 @@ message CheckinAgentInfo {
     bool snapshot = 3;
 }
 
+// Feature flags configurations.
+// Added on Elastic Agent v8.7.1.
+message Features {
+    FQDNFeature fqdn = 2;
```
blakerouse, line 298: I would also add the `Source` field just like the UnitExpectedConfig does. That will allow future features to be added and used without requiring the client to be modified and updated.

pkg/client/client_v2.go
```diff
@@ -22,35 +22,56 @@ import (
...
+//   - a unit is added, modified, or removed
+//   - a feature changes
 type UnitChanged struct {
-	Type UnitChangedType
+	Type     ChangeType
+	Triggers []Trigger
+	// Unit is any change in a unit.
 	Unit *Unit
+	// Features are all the feature flags and their configs.
+	Features Features
```
blakerouse, line 74: Don't add Features here, place it inside the `Expected()` result inside the unit. I am also okay with changing the return type of `Expected()` to be a struct.

pkg/client/client_v2.go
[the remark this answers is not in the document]
blakerouse: No need to change this name.

pkg/client/client_v2.go
```diff
@@ -118,11 +139,13 @@ type clientV2 struct {
...
-	kickCh  chan struct{}
-	errCh   chan error
-	unitsCh chan UnitChanged
-	unitsMu sync.RWMutex
-	units   []*Unit
+	kickSendObservedCh chan struct{}
+	errCh              chan error
+	changesCh          chan UnitChanged
+	featuresMu         sync.Mutex
+	features           Features
```
blakerouse, line 146: Going to need to add the `featuresIdx` here as well. I would start the initial index at 1.

There is a reason for starting it at 1, and that is backwards compatibility. With older clients, I would modify the Elastic Agent to read the observed features idx at 0, as the client doesn't understand features and we should not trigger an `unsettle` when it doesn't match, because it will never match. If it gets a 1, then it knows its a client that supports it but doesn't know about it. Or you add a boolean field in the observed message to signal that the client supports features.

magefile.go
```diff
@@ -54,9 +56,14 @@ func (Prepare) InstallGoLint() error {
...
-// All runs prepare:installGoLicenser and prepare:installGoLint.
+// InstallStringer install go stringer to generate String methods for constants.
+func (Prepare) InstallStringer() error {
+	return GoInstall(stringerRepo)
+}
+
+// All runs prepare:installGoLicenser, prepare:installGoLint and prepare:installGoLint.
 func (Prepare) All() {
-	mg.SerialDeps(Prepare.InstallGoLicenser, Prepare.InstallGoLint)
+	mg.SerialDeps(Prepare.InstallGoLicenser, Prepare.InstallGoLint, Prepare.InstallStringer)
```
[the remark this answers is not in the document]
ycombinator: Looks like stringer usage has been removed.  So I'm going to go ahead and remove the related changes from `magefile.go` as well.

ycombinator:
There are a number of cleanup changes in this PR that aren't strictly related to the feature flag implementation.  To make it easier to review this PR (for selfish reasons 😄), I've extracted the cleanup changes into their own PR: https://github.com/elastic/elastic-agent-client/pull/61.  

Once that cleanup PR is merged, I'll rebase this one on `main` so it ends up containing only FQDN-related changes.

pkg/client/client_v2.go
ycombinator: Unless I'm missing something, I'm not sure we need `features` (and related fields) here on the `clientV2` struct at all, since `features` is a field on the `Unit` struct (similar to `logLevel`).

pkg/client/client_v2.go
```diff
@@ -22,19 +23,72 @@ import (
...
 )
 
-// String returns string representation for the unit changed type.
+const (
+	// TriggeredNothing is Trigger zero value, nothing was triggered.
+	// It exists only for completeness and documentation purposes.
+	TriggeredNothing Trigger = 0 // nothing_triggered
+
+	// TriggeredConfigChange indicates a change in config triggered the change.
+	TriggeredConfigChange Trigger = 1 << iota // config_change_triggered
```
michalpristas, line 48: can you leave a comment that event can have multiple triggers and this is a mask so nobody accidentally changes this

ycombinator: Done.

pkg/client/unit.go
```diff
@@ -146,11 +151,25 @@ func (u *Unit) Type() UnitType {
...
+// Expected contains the expected state, log level, features and config for a unit.
+type Expected struct {
+	Config   *proto.UnitExpectedConfig
+	Features *proto.Features
+	LogLevel UnitLogLevel
+	State    UnitState
+}
+
+// Expected returns the expected state, log level, features and config for the unit.
+func (u *Unit) Expected() Expected {
```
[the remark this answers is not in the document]
ycombinator: I could be misunderstanding I'm still somewhat new to the Agent codebase, but I'm not sure much has changed from a "contract" perspective here?  It looks like, before this PR `Expected()` was returning three values whereas it's returning four values now, but encapsulated in an `Expected` struct object.

ycombinator:
> alex had a good comment about Expected though

++ addressed it in the same thread; tl;dr: see https://github.com/elastic/elastic-agent-client/issues/62.

PR #63, opened 2023-03-11
Send feature flags information as part of observed state in the next checkin

ycombinator:
Follow up to https://github.com/elastic/elastic-agent-client/pull/56.

While updating the `management.TestManagerV2` test in https://github.com/elastic/beats/pull/34456 to account for feature flags, I realized `elastic-agent-client` wasn't reporting it's current feature flag status in the observed state sent to the server (Agent) during checkin.  This PR fixes that miss.

### DIFF
--- a/pkg/client/client_v2.go
+++ b/pkg/client/client_v2.go
@@ -194,7 +194,6 @@ type clientV2 struct {
 	units              []*Unit
 
 	featuresMu  sync.RWMutex
-	features    *proto.Features
 	featuresIdx uint64
 
 	dmx       sync.RWMutex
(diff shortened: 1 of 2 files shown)


pkg/client/client_v2.go
```diff
@@ -496,11 +506,21 @@ func (c *clientV2) syncUnits(expected *proto.CheckinExpected) {
 		}
 	}
 
+	// Now that we've propagated feature flags' information to units, record
+	// the same information on the client so we can send it up as part of the
+	// observed on c so we can send up as part of the observed state in the next
+	// checkin.
```
cmacknz, line 512:
```suggestion
	// the same information on the client so we can send it up as part of the observed state in the next checkin.
```

cmacknz (approved):
Looks good. Is there a unit test we can add here for this? I know we catch it on the Beats side now, but ideally we'd catch it at the source in this repository.

ycombinator:
@cmacknz said:

> Is there a unit test we can add here for this? 

Unit test added.

I made another change to this PR since your last review.  After chatting with @blakerouse, I changed the expected checkin message to only contain `FeaturesIdx`, not `Features` as well.

blakerouse (approved):
This looks good!

PR #64, opened 2023-03-27
Move deprecated and future use messages into separate files.

cmacknz:
Make the elastic-agent-client.proto file easier to read by moving the deprecated checkin V1 messages and the unimplemented future services into their own files. 

I also moved the service definitions to come before the messages definitions to make the files easier to search when reading but left the order of message definition unchanged.

The primary elastic-agent-client.proto file now only contains messages that exist and are used today.

I could take this further and delete the deprecated and future files entirely instead preserving them on a branch if we wanted. This would probably be my preference, but just using separate files is less drastic.


cmacknz:
Here's the resulting elastic-agent-client.proto file without all the noise in the diff from moving code around: https://github.com/cmacknz/elastic-agent-client/blob/separate-protocol-versions/elastic-agent-client.proto

michalpristas:
a lot of removal, have you tested this with upgrade? (scenario upgrading to agent with this client from stale one)

ycombinator (commented):
The changes make sense to me.

It would be nice to know that the API exposed by the `github.com/elastic/elastic-agent-client/pkg/proto` package didn't change with these changes.  I've been looking for some AST-based package comparison tools but haven't found anything yet.

cmacknz:
Built an 8.8.0 agent with this change, enrolled it with Fleet, and everything works fine.

ycombinator (approved):
LGTM.

PR #69, opened 2023-04-26
fix diagnostics pprof output

leehinman:
- changed to non-debug output
- add test to verify output is valid pprof
- change content-type
- change filename suffix to pprof.gz

Relates elastic/elastic-agent#2530

### DIFF
--- a/pkg/client/client_v2.go
+++ b/pkg/client/client_v2.go
@@ -516,7 +516,6 @@ func (c *clientV2) syncUnits(expected *proto.CheckinExpected) {
 		// otherwise it will not be notified until the next checkin timeout
 		c.unitChanged()
 	}
-
 }
 
 // findUnit finds an existing unit.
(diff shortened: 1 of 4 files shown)


pkg/client/client_v2.go
```diff
@@ -781,18 +780,29 @@ func (c *clientV2) tryPerformDiagnostics(actionResults chan *proto.ActionRespons
...
+	c.registerPprofDiagnostics("heap", "a sampling of memory allocations of live objects", 0)
+	c.registerPprofDiagnostics("allocs", "a sampling of all past memory allocations", 0)
+	c.registerPprofDiagnostics("threadcreate", "stack traces that led to the creation of new OS threads", 0)
+	c.registerPprofDiagnostics("block", "stack traces that led to blocking on synchronization primitives", 0)
+	c.registerPprofDiagnostics("mutex", "stack traces of holders of contended mutexes", 0)
 }
 
-func (c *clientV2) registerPprofDiagnostics(name string, description string) {
-	c.RegisterDiagnosticHook(name, description, fmt.Sprintf("%s.txt", name), "plain/text", func() []byte {
+func (c *clientV2) registerPprofDiagnostics(name string, description string, debug int) {
```
cmacknz, line 791: The agent depends on elastic-agent-client, maybe we could keep this implementation here so it isn't duplicated from https://github.com/elastic/elastic-agent-client/pull/69/?

cmacknz (approved):
Changes looks as good it did the first time. Nit that we've duplicated this code though.

Next up:
Updating this dependency in Beats and Fleet Server

cmacknz:
@norrietaylor @florianl you'll want to update your dependencies on elastic-agent-client to pick up this change, we were capturing heap profiles in the wrong format in our default set of diagnostics handlers.

Issue #66, opened 2023-04-05
Client checkin spinlocks on RPC error


faec:
In `client/client.go`, `(*client).startCheckin` has a loop calling through to `(*client).checkinRoundTrip`. In normal operation the rate of these round trips is gated by the pace of the Agent's checkin responses, however if there is an error condition that prevents connecting to the Agent the loop will fall through immediately, spinlocking on the error. This manifests in live deployments as e.g. a gRPC `PermissionDenied` error occurring 5-10K times per second, which both worsens the effects of the error and floods the logs so that any clues about the original cause of the error are lost.

This error loop should have some minimum retry interval to prevent this symptom.

PR #70, opened 2023-05-17
Fix RPC error spinlock / clean up clientV2

faec:
Add a minimum timeout to `(*clientV2).startCheckin` and `(*clientV2).startActions` so retries of `checkinRoundTrip` and `actionRoundTrip` don't spinlock if there's an RPC error (see https://github.com/elastic/elastic-agent-client/issues/66).

Fixes https://github.com/elastic/elastic-agent-client/issues/66.
fix https://github.com/elastic/elastic-agent/issues/2628

At the same time, break some monolithic functions up into multiple helpers, fix some race conditions, simplify the synchronization variables, and add explanatory comments for many steps of the checkin process.

### DIFF
--- a/pkg/client/client_v2.go
+++ b/pkg/client/client_v2.go
@@ -290,6 +290,20 @@ func (c *clientV2) RegisterDiagnosticHook(name string, description string, filen
 	}
 }
 
+// durationOf wraps a given function call and measures the time it consumed.
+func durationOf(callback func()) time.Duration {
+	before := time.Now()
+	callback()
+	return time.Since(before)
+}
+
+func nextExpBackoff(current time.Duration, max time.Duration) time.Duration {
+	if current*2 >= max {
+		return max
+	}
+	return current * 2
+}
+
 // startCheckin starts the go routines to send and receive check-ins
 //
 // This starts 3 go routines to manage the check-in bi-directional stream. The first
@@ -305,9 +319,18 @@ func (c *clientV2) startCheckin() {
 		// If the initial RPC connection fails, checkinRoundTrip
 		// returns immediately, so we set a timer to avoid spinlocking
 		// on the error.
-		retryTimer := time.NewTimer(1 * time.Second)
+		retryInterval := 1 * time.Second
+		retryTimer := time.NewTimer(retryInterval)
 		for c.ctx.Err() == nil {
-			c.checkinRoundTrip()
+			checkinDuration := durationOf(c.checkinRoundTrip)
+
+			if checkinDuration < time.Second {
+				// If the checkin round trip lasted less than a second, we're
+				// probably in an error loop -- apply exponential backoff up to 30s.
+				retryInterval = nextExpBackoff(retryInterval, 30*time.Second)
+			} else {
+				retryInterval = 1 * time.Second
+			}
 
 			// After the RPC client closes, wait until either the timer interval
 			// expires or the context is cancelled. Note that the timer waits
@@ -316,7 +339,7 @@ func (c *clientV2) startCheckin() {
 			// connection shuts down.
 			select {
 			case <-retryTimer.C:
-				retryTimer.Reset(1 * time.Second)
+				retryTimer.Reset(retryInterval)
 			case <-c.ctx.Done():
 			}
 		}
@@ -573,9 +596,20 @@ func (c *clientV2) startActions() {
 		// If the initial RPC connection fails, actionRoundTrip
 		// returns immediately, so we set a timer to avoid spinlocking
 		// on the error.
-		retryTimer := time.NewTimer(1 * time.Second)
+		retryInterval := 1 * time.Second
+		retryTimer := time.NewTimer(retryInterval)
 		for c.ctx.Err() == nil {
-			c.actionRoundTrip(actionResults)
+			actionDuration := durationOf(func() {
+				c.actionRoundTrip(actionResults)
+			})
+
+			if actionDuration < time.Second {
+				// If the action round trip lasted less than a second, we're
+				// probably in an error loop -- apply exponential backoff up to 30s.
+				retryInterval = nextExpBackoff(retryInterval, 30*time.Second)
+			} else {
+				retryInterval = 1 * time.Second
+			}
 
 			// After the RPC client closes, wait until either the timer interval
 			// expires or the context is cancelled. Note that the timer waits
@@ -584,7 +618,7 @@ func (c *clientV2) startActions() {
 			// connection shuts down.
 			select {
 			case <-retryTimer.C:
-				retryTimer.Reset(1 * time.Second)
+				retryTimer.Reset(retryInterval)
 			case <-c.ctx.Done():
 			}
 		}
--- a/pkg/client/client_v2.go
+++ b/pkg/client/client_v2.go
@@ -177,23 +177,25 @@ type clientV2 struct {
 	agentInfoMu sync.RWMutex
 	agentInfo   *AgentInfo
 
-	versionInfo     VersionInfo
-	versionInfoSent bool
+	versionInfo         VersionInfo
+	sendVersionInfoOnce sync.Once
 
 	ctx    context.Context
 	cancel context.CancelFunc
 	wg     sync.WaitGroup
 	client proto.ElasticAgentClient
-	cfgMu  sync.RWMutex
-	obsMu  sync.RWMutex
 
-	kickSendObservedCh chan struct{}
-	errCh              chan error
-	changesCh          chan UnitChanged
-	unitsMu            sync.RWMutex
-	units              []*Unit
+	errCh     chan error
+	changesCh chan UnitChanged
 
-	featuresMu  sync.RWMutex
+	// stateChangeObservedCh is an internal channel that notifies checkinWriter
+	// that a unit state has changed. To trigger it, call unitsStateChanged.
+	stateChangeObservedCh chan struct{}
+
+	unitsMu sync.RWMutex
+	// Must hold unitsMu to access units or featuresIdx, since they
+	// must be synchronized with each other.
+	units       []*Unit
 	featuresIdx uint64
 
 	dmx       sync.RWMutex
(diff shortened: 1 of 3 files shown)


blakerouse:
@faec Description says unit test was added, don't see it in the changes, was it not committed?

pkg/client/client_v2.go
```diff
@@ -300,130 +302,129 @@ func (c *clientV2) startCheckin() {
...
 
+func (c *clientV2) checkinWriter(
+	checkinClient proto.ElasticAgent_CheckinV2Client,
+	done chan struct{},
+) {
+	t := time.NewTicker(c.minCheckTimeout)
+	defer t.Stop()
+
+	// Keep sending until the call returns an error
+	for c.sendObserved(checkinClient) == nil {
```
cmacknz, line 333: Another option that doesn't rely on us manually wrapping each call is to look at using gRPC [interceptors](https://github.com/grpc/grpc-go/blob/master/examples/features/interceptor/README.md#interceptor) to do this to inject the retries into each underlying gRPC call.

Here's an implementation of a retry interceptor I've used before with linear and exponential backoff support:
https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/v2/interceptors/retry

faec:
> @faec Description says unit test was added, don't see it in the changes, was it not committed?

Yes, it's there now, I didn't want to push it until it actually ran successfully. The new test passes with this PR, and against the base branch produces:

```
--- FAIL: TestRPCErrorRetryTimer (0.05s)
    client_v2_test.go:73:
        	Error Trace:	client_v2_test.go:73
        	Error:      	Not equal:
        	            	expected: 2
        	            	actual  : 15763
        	Test:       	TestRPCErrorRetryTimer
        	Messages:   	expected 2 RPC errors, one from checkinRoundTrip and one from actionRoundTrip
```

pkg/client/client_v2.go
faec: This sounds nice, but I just tried it and it doesn't seem to be working -- I've pushed my attempt so you can see it and suggest if there's anything I missed, but running it produces:

```
--- FAIL: TestRPCErrorRetryTimer (0.05s)
    client_v2_test.go:73:
        	Error Trace:	client_v2_test.go:73
        	Error:      	Not equal:
        	            	expected: 2
        	            	actual  : 11667
        	Test:       	TestRPCErrorRetryTimer
        	Messages:   	expected 2 RPC errors, one from checkinRoundTrip and one from actionRoundTrip
```

I tried a few different ways of setting it and can't find anything that works, so I think my theory is that this wraps retries of _calls_ to the RPC host, not the initial connection, which is what's causing the bug. (This fits with the interceptor docs saying it wraps "around the user call" which plausibly wouldn't include the creation of the client itself.) Let me know if you have other ideas, otherwise I can roll back to the previous implementation.

faec:
(Current pushed version doesn't pass tests, see the conversation thread with @cmacknz about alternate implementations. The working version is https://github.com/elastic/elastic-agent-client/pull/70/files/11cfec3a29e09a5e4435d285865f93c6059fd096, which is what I'll roll back to if we can't find a workaround for the suggested helper library.)

pkg/client/client_v2.go
cmacknz: Ah yes, it is retrying the individual calls and not the initial connection.

pierrehilbert:
@jlind23 from my understanding we shouldn't, the root cause for this issue could be https://github.com/elastic/elastic-agent/issues/2691 and we should wait to confirm that after both we fixed the issue.

faec:
@pierrehilbert @jlind23 It looks to me like this does fix https://github.com/elastic/elastic-agent/issues/2628 -- the deadlock in https://github.com/elastic/elastic-agent/issues/2691 is one way to trigger the log spamming this PR addresses, but there are others (we have also seen it happen with other types of RPC error). This PR doesn't fix the deadlock, which has its own issue, but it does fix the log spamming, so even if the deadlock happens, we'll still get usable logs.

pkg/client/client_v2.go
```diff
@@ -317,113 +327,105 @@ func (c *clientV2) checkinRoundTrip() {
...
 	msg := &proto.CheckinObserved{
 		Token:       c.token,
 		Units:       observed,
-		FeaturesIdx: c.featuresIdx,
+		FeaturesIdx: featuresIdx,
 		VersionInfo: nil,
 	}
-	if !c.versionInfoSent {
-		c.versionInfoSent = true
+	c.sendVersionInfoOnce.Do(func() {
```
cmacknz, line 413: This only guarantees that we have filled the `VersionInfo` struct once, not that it has been successfully sent. It would be more robust to stop doing this once the `Send` call below succeeds when `VersionInfo` is set.

It looks like this was a problem before as well.

pkg/client/client_v2.go
```diff
@@ -528,10 +540,15 @@ func (c *clientV2) findUnit(id string, unitType UnitType) *Unit {
 	return nil
 }
 
-// unitChanged triggers the send goroutine to send a new observed state
-func (c *clientV2) unitChanged() {
-	if len(c.kickSendObservedCh) <= 0 {
-		c.kickSendObservedCh <- struct{}{}
+// unitsStateChanged notifies checkinWriter that there is new
+// state data to send.
+func (c *clientV2) unitsStateChanged() {
```
cmacknz, line 545: I prefer the old name `unitChange`, since the unit itself contains a `State` field and I believe we want to call this when anything in the unit changes not just the state.

pkg/client/client_v2.go
```diff
@@ -551,102 +568,118 @@ func (c *clientV2) startActions() {
...
+			c.actionRoundTrip(actionResults)
+
+			// After the RPC client closes, wait until either the timer interval
+			// expires or the context is cancelled. Note that the timer waits
+			// one second since the checkinRoundTrip was last _called_, not
+			// since it returns; immediate retries are ok after an active
+			// connection shuts down.
 			select {
+			case <-retryTimer.C:
+				retryTimer.Reset(1 * time.Second)
```
cmacknz, line 585: Would it be worth using an exponential backoff here and in the checkinV2 call?

At one hand we want to reconnect to the agent relatively quickly for transient errors, on the other hand permanent failures like the deadlocks we keep seeing are just going to see the logs filled with these 1s interval error message eventually.

cmacknz:
The changes here look reasonable to me. Before merging this let's do the following if you haven't already:

1. Package Beats with this version of the Elastic Agent client (just Metricbeat and Filebeat are enough).
2. Package the agent using `EXTERNAL=false` so that it will use the version of Beats you just built.
3. Enroll the agent in Fleet and ensure it is healthy and:
  i. Verify changing the log level works.
  ii. Verify that uninstalling and re-installing the system integration works and the agent stays healthy.
  iii. Add an additional output parameter like `bulk_max_size` and ensure it applies and the agent stays healthy.

That should be enough to sanity check we haven't accidentally broken anything major. It will change inputs, outputs, and send the log level change action. The changes look fine but the code here has a history of hard to detect problems.

pkg/client/client_v2.go
faec: Fwiw this is only called in two places:
- In `syncUnits`, when a unit is removed
- In `(*Unit).UpdateState`, when a new `UnitState` is received. The fields that can be modified when this happens are `state`, `stateMsg`, and `statePayload`.

So I think in this case "state" is fairly accurate -- most changes to the `Unit` structure itself will not trigger this reload (because the state is the only part that needs to be reported back to Agent). But if you still prefer the other name after this explanation I can change it :-)

pkg/client/client_v2.go
faec: Ah, good catch. And correct me if I'm wrong but it looks like it's important that we send it at least once but not very important that we never send it more than once? So I've reverted to a non-synchronized variable but moved its update to after the Send.

pkg/client/client_v2.go
cmacknz: Re-reading the code I think its fine as is, thanks.

pkg/client/client_v2.go
faec: I thought about it during implementation and decided it was overkill but I will poke at it again during testing...

faec:
> Before merging this let's do the following if you haven't already:...

I had been expecting to rely on the unit tests for this merge and do larger-scale testing when updating the Beats repo, but you're right, caution is a good idea, I will follow up on this after on week.

pkg/client/client_v2.go
cmacknz: Yes it needs to get through at least once, afterwards not sending it as an optimization to avoid pointless copying bytes that will never change between processes.

Here's the agent side of this where it pulls the values out:

https://github.com/elastic/elastic-agent/blob/78774a78a4a758bfd1c7a2a6f95fbed66c393d92/pkg/component/runtime/state.go#L322-L335

```go
	if checkin.VersionInfo != nil {
		if checkin.VersionInfo.Name != "" && s.VersionInfo.Name != checkin.VersionInfo.Name {
			s.VersionInfo.Name = checkin.VersionInfo.Name
			changed = true
		}
		if checkin.VersionInfo.Version != "" && s.VersionInfo.Version != checkin.VersionInfo.Version {
			s.VersionInfo.Version = checkin.VersionInfo.Version
			changed = true
		}
		if checkin.VersionInfo.Meta != nil && diffMeta(s.VersionInfo.Meta, checkin.VersionInfo.Meta) {
			s.VersionInfo.Meta = checkin.VersionInfo.Meta
			changed = true
		}
	}
```

pkg/client/client_v2.go
```diff
@@ -317,113 +327,107 @@ func (c *clientV2) checkinRoundTrip() {
...
+// forwards it to the client's error channel before returning.
 func (c *clientV2) sendObserved(client proto.ElasticAgent_CheckinV2Client) error {
 	c.unitsMu.RLock()
 	observed := make([]*proto.UnitObserved, 0, len(c.units))
 	for _, unit := range c.units {
 		observed = append(observed, unit.toObserved())
 	}
+	// Check featuresIdx within the same mutex block since we
+	// need observed and featuresIdx to be synchronized with each other.
+	featuresIdx := c.featuresIdx
```
michalpristas, line 404: 👍

pkg/client/client_v2.go
```diff
@@ -437,26 +441,42 @@ func (c *clientV2) sync(expected *proto.CheckinExpected) {
...
 }
 
-func (c *clientV2) syncUnits(expected *proto.CheckinExpected) {
-	c.unitsMu.Lock()
-	defer c.unitsMu.Unlock()
-	i := 0
-	removed := false
+// Remove from the client any units not in the given expected list,
+// returning a count of the units removed.
+// Caller must hold c.unitsMu.
```
michalpristas, line 446: thanks for adding a warning

michalpristas (approved):
I like how much more readable it is especially I like removal of all the nesting

cmacknz (approved):
All my comments were addressed, but forgot to approve. Thanks!

PR #74, opened 2023-07-03
Migration to Buildkite

leo-ri:


.buildkite/pipeline.yml
```diff
@@ -1,5 +1,44 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+env:
+  SETUP_GOLANG_VERSION: '1.18.10'
```
mrodm, line 3: It's a pitty there is no `.go-version` file in this repository.

.buildkite/scripts/update.sh
```diff
@@ -0,0 +1,18 @@
...
+
+source .buildkite/scripts/pre-install-command.sh
+apt-get update
+apt-get install unzip
+curl -sSfL -o protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v3.19.4/protoc-3.19.4-linux-x86_64.zip
+unzip -o protoc.zip -d /usr/local
+which protoc
+
+go env
+echo "$PATH"
```
mrodm, line 13: Is this for debugging ? 
Could it be removed ?

.buildkite/pipeline.yml
```diff
@@ -1,5 +1,44 @@
...
 
 steps:
-  - label: "Example test"
-    command: echo "Hello!"
+  - label: "Update"
+    key: update
+    command:
+      - ".buildkite/scripts/update.sh"
```
mrodm, line 9: It looks like that the Jenkins pipeline is intended to be run everything in the same agent
https://github.com/elastic/elastic-agent-client/blob/aa961fb53dddc0aadff1858582558967222eecb6/.ci/Jenkinsfile#L6

Should these steps be run in the same buildkite agent too ? 
Do lint or Test require what this steps generates (GRPC code) ?

.buildkite/pipeline.yml
leo-ri: means to add it?

mrodm: No, if they are not using it. Let's keep it as it was set in Jenkins ([using an env. variable](https://github.com/elastic/elastic-agent-client/blob/aa961fb53dddc0aadff1858582558967222eecb6/.ci/Jenkinsfile#L10)).

PR #75, opened 2023-07-03
Allow dial options when initializing a V2 client from reader

dmathieu:
I want to enable gRPC interceptors with this client, so we can get telemetry/tracing with them.
However, while `NewV2` accepts a set of DialOptions, NewV2FromReader, which calls the same method under the hood doesn't.


blakerouse (approved):
Looks good!

PR #82, opened 2023-08-10
Add apm config propagation

pchila:
This change allows for agent to propagate APM configuration to components.
The only APM configuration currently supported is Elastic APM but more can be added in the future.


elastic-agent-client.proto
```diff
@@ -185,9 +185,32 @@ message FQDNFeature {
     bool enabled = 1;
 }
 
+// Elastic APM TLS config
+message ElasticAPMTLS {
```
ycombinator, line 189: Would it make sense to make this is more generic `TLS` message?

pkg/client/client_v2.go
```diff
@@ -568,6 +578,15 @@ func (c *clientV2) syncUnits(expected *proto.CheckinExpected) {
...
 	}
 
+	var apmConfig *proto.APMConfig
+	func() {
+		c.componentMu.RLock()
+		defer c.componentMu.RUnlock()
+		if c.componentConfig != nil {
+			apmConfig = c.componentConfig.ApmConfig
+		}
+	}()
```
ycombinator, line 588: Why use an IIFE here? Is it because of the `defer`?

pkg/client/client_v2_test.go
```diff
@@ -869,6 +875,149 @@ func TestClientV2_Checkin_FeatureFlags(t *testing.T) {
...
+					uca.unitChanges = append(uca.unitChanges, uc)
+				}()
+			}
+		}
+	}()
+}
+
+func (uca *unitChangesAccumulator) GetAccumulatedChanges() []UnitChanged {
+	uca.mx.Lock()
+	defer uca.mx.Unlock()
```
ycombinator, line 912: Could these be `uca.mx.RLock()` and `uca.mx.RUnlock()` with `uc.mx` becoming an `RWMutex`?

pkg/client/client_v2_test.go
```diff
@@ -869,6 +875,149 @@ func TestClientV2_Checkin_FeatureFlags(t *testing.T) {
...
+
+	ret := make([]UnitChanged, len(uca.unitChanges))
+	copy(ret, uca.unitChanges)
```
ycombinator, line 915: Not sure how much this matters but if `uca.unitChanges` is `nil`, `ret` will not be `nil` but instead be a slice of length 0. See https://go.dev/play/p/jzJ9cXkR6pZ.

pkg/client/unit_test.go
```diff
@@ -59,21 +59,109 @@ func TestUnitUpdateWithNewMap(t *testing.T) {
...
+	type testcase struct {
+		name         string
+		initialState Unit
+		update       struct {
+			//unitState   UnitState
+			//UnitLogLvl  UnitLogLevel
+			//featuresIdx uint64
+			//features    *proto.Features
+			//unitConfig  *proto.UnitExpectedConfig
+			//cfgIdx      uint64
```
ycombinator, line 116: Did you intentionally leave these commented-out fields in?

elastic-agent-client.proto
pchila: Not sure if a generic TLS message will have the same format so I kept it only for Elastic APM in this case...

elastic-agent-client.proto
```diff
@@ -185,9 +185,32 @@ message FQDNFeature {
...
+
+// APM configuration
+message APMConfig {
+    ElasticAPM elastic = 1;
+}
+
 // Component-level configuration.
 message Component {
     ComponentLimits limits = 1;
+    optional APMConfig apm_config = 2;
```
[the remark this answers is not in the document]
pchila: In the original issue it was already discussed sending OTel config to the components directly with the Elastic APM config being the first iteration.
In my mind that means that the OTel and Elastic APM config will live side by side for a while and then the ElasticAPM would finally be retired that's why there's the extra indirection of APM config...

For a more general case I could easily see a configuration where multiple APM instrumentation are used across different components and maybe we want to give multiple configs for different protocols (even if probably OTel could take care of that scenario in a few years if it has a big adoption)

pkg/client/client_v2.go
pchila: it is to guarantee that whatever happens we unlock, yes

pkg/client/client_v2_test.go
pchila: In the test there is exactly 1 writer (the accumulator) and 1 reader (the test itself): an RWMutex would make sense in case we have multiple readers and (maybe) multiple writers as we allow concurrent read access reducing contention...

For this case I think that a simple mutex is enough (we are not looking for performance or anything, just avoiding data races)

pkg/client/client_v2_test.go
pchila: It does not matter too much here since we are going to check the length and contents and in both cases we have 0 length (so no content)
I don't think that the extra allocation for the zero-length slice is gonna cost us here...

pkg/client/unit_test.go
pchila: Ah this was a quick copy paste that I used to remind myself what could go in a unit object in case I needed to set multiple fields in the test. 
In the end I only needed the apm config so everything else remained commented, I will remove the extrase

AndersonQ (approved):

pkg/client/client_v2.go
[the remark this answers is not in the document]
pchila: This is for safety as we are guaranteed to unlock whatever happens in the function

pkg/client/unit.go
```diff
@@ -326,6 +331,10 @@ func (u *Unit) updateState(
 		}
 	}
 
+	if ((u.apm == nil || expAPM == nil) && u.apm != expAPM) || !gproto.Equal(u.apm, expAPM) {
```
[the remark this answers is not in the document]
pchila: Yes it can :smile:

PR #83, opened 2023-08-11
Add component-level configuration with go_max_procs limit

rdner:
`CheckinExpected`/`CheckinObserved` now has the `component` field with a component-level parameters. One of them is `limits` which contains the `go_max_procs` that defines how many CPUs the component is allowed to use in a Go runtime.

This value is handled by the Go implementation of the client setting `runtime.GOMAXPROCS` to `ComponentLimits.go_max_procs`.

Non-Go clients should ignore this value.


elastic-agent-client.proto
```diff
@@ -185,6 +185,24 @@ message FQDNFeature {
...
+
+// Limits to configure for the currently running component.
+message ComponentLimits {
+  // Source is the original source of the limits. All values from the limits
+  // are included here even if other concrete fields are defined for this message.
+  google.protobuf.Struct source = 1;
+
+  // The maximum number of CPUs that can be used simultaneously by the component.
+  // For Go programs, it's GOMAXPROCS, see https://pkg.go.dev/runtime#GOMAXPROCS.
+  // Other languages should set their equivalent to Go's GOMAXPROCS.
```
cmacknz, line 201: As I mentioned in https://github.com/elastic/ingest-dev/issues/1656#issuecomment-1673670217 this pretty much doesn't exist and I would prefer we don't enforce this as an obligation on input/component developers.

If we can find a well know example that isn't Go we can keep the generic name, to avoid confusion I would prefer to specifically name this `go_max_procs`.

cmacknz: This limit is highly specific to the Go threading model where it is implementing its own scheduler. Languages without this default to the OS level thread scheduler and controlling this isn't available from the language itself.

pkg/client/client_v2.go
```diff
@@ -510,6 +520,24 @@ func (c *clientV2) removeUnexpectedUnits(expected *proto.CheckinExpected) int {
...
+func (c *clientV2) syncComponent(expected *proto.CheckinExpected) {
+	if expected.Component == nil {
+		return
+	}
+	c.componentMu.Lock()
+	defer c.componentMu.Unlock()
+
+	if expected.Component.Limits.MaxProcs > 0 {
+		_ = runtime.GOMAXPROCS(int(expected.Component.Limits.MaxProcs))
+	}
```
cmacknz: If MaxProcs is set to a value < runtime.NumCPU and then the setting is removed the policy, this will just leave the setting as the previous value.

I think it would be much more convenient if removing the setting from the policy or setting it to zero returned it to the default value of runtime.NumCPU. 

Otherwise a user has no way to go back to the default setting because there is no guarantee that all the agents in a policy are executing on machines with the same number of CPUs. There is no one value they can set to go back to the number of CPUs on the machine.


This call also may not be free from a performance perspective, it locks the global scheduler lock on each call. The implementation is in https://cs.opensource.google/go/go/+/refs/tags/go1.21.0:src/runtime/debug.go;l=16

```go
func GOMAXPROCS(n int) int {
	if GOARCH == "wasm" && n > 1 {
		n = 1 // WebAssembly has no threads yet, so only one CPU is possible.
	}

	lock(&sched.lock)
	ret := int(gomaxprocs)
	unlock(&sched.lock)
	if n <= 0 || n == ret {
		return ret
	}

	stopTheWorldGC(stwGOMAXPROCS)

	// newprocs will be processed by startTheWorld
	newprocs = int32(n)

	startTheWorldGC()
	return ret
}
```

It may be best to only call this function when we have to. It should be cheap for us to save the most recently applied value and skip this call if it isn't needed.

pchila (approved):
LGTM

pkg/client/client_v2.go
```diff
@@ -529,12 +531,28 @@ func (c *clientV2) syncComponent(expected *proto.CheckinExpected) {
...
 		return
 	}
+
 	c.componentMu.Lock()
 	defer c.componentMu.Unlock()
 
-	if expected.Component.Limits.MaxProcs > 0 {
-		_ = runtime.GOMAXPROCS(int(expected.Component.Limits.MaxProcs))
+	// applying the component limits
+	if expected.Component.Limits != nil {
```
cmacknz, line 539: This logic could have a small unit test, otherwise looks good.

cmacknz (approved):

pkg/client/client_v2.go
rdner: Yeah, that's what I've been doing for the last few hours :) This code was buggy, I added tests and fixed the bugs in https://github.com/elastic/elastic-agent-client/pull/83/commits/52a5535723c05c45867ff66b9b5fb75c1e7b4008

PR #84, opened 2023-08-15
Add a README explaining the protocol design and include example messages

cmacknz:
Add a README file and explain the message fields in more detail. Explain the design of the protocol and add more details to each individual RPC call. In particular document the index fields in more depth, explaining what happens when a unit is removed, and add a sample message exchange between a Beat and the agent.

To see the rendered README file go to https://github.com/cmacknz/elastic-agent-client/tree/document-control-protocol


elastic-agent-client.proto
```diff
@@ -268,14 +273,14 @@ message CheckinObserved {
     // one of the values have changed.
     optional CheckinObservedVersionInfo version_info = 3;
 
-    // Features are the expected feature flags configurations.
-    // Added on Elastic Agent v8.7.1.
-    Features features = 4;
```
cmacknz, line 273: This deletion is backwards compatible, it isn't even sent in the messaging today.

README.md
```diff
@@ -0,0 +1,418 @@
...
+              "version": "1.38.2"
+            }
+          },
+          "dataStream": {
+            "source": {
+              "namespace": "default"
+            },
+            "namespace": "default"
+          },
+          "streams": [
```
ycombinator, line 204: What is the difference between .expected.units[n].config.source.streams` and `.expected.units[n].config.streams`?

README.md
```diff
@@ -0,0 +1,418 @@
...
+    var expected *proto.CheckinExpected,
+    var observed *proto.CheckinObserved,
+
+    obsJSON, _ := protojson.Marshal(observed)
+    expJSON, _ := protojson.Marshal(expected)
+    c.logger.Infow("Checkin",
+        zap.Any("observed", json.RawMessage(obsJSON)),
+        zap.Any("expected", json.RawMessage(expJSON)),
+    )
+```
```
ycombinator, line 418: Would it be possible to include this code in Agents built with `DEV=true`?

README.md
cmacknz: I need to add a top level explanation of the source field.

The source is the entire contents of the input section of the policy unmodified. the `.expected.units[n].config.streams` is only what matched against the proto message definition in this file which may not be the entire configuration. We only pull out the standardized/mandatory parts of it.

There is an attempt to explain this in the message but this pattern is used a lot, I will add a higher level explanation and make a specific note in this section.

https://github.com/elastic/elastic-agent-client/blob/c699c976fa3092435985dd633c1ed7807a753e74/elastic-agent-client.proto#L125-L142

elastic-agent-client.proto
```diff
@@ -10,34 +10,29 @@ option cc_enable_arenas = true;
...
     // Called by the client to provide the Elastic Agent the state of the application over the V2 protocol.
     //
-    // A `CheckinObserved` must be streamed at least every 30 seconds or it will result in the
-    // set of units automatically marked as FAILED, and after 60 seconds the Elastic Agent will
+    // Implements a reconciliation loop where a component periodically tells the agent what its current
+    // observed configuration is, and the agent replies with the configuration it is expected to be running.
+    //
+    // Each included configuration block is accompanied by an index or revision number so that the
+    // observed messages do not need to waste CPU copying the entire applied configuration back to
+    // the agent on each checkin. Configurations in large deployments can be 1MB or more.
```
ycombinator, line 22:
```suggestion
    // Each configuration block included in the expected message is accompanied by an index or revision number. Corresponding 
    // observed messages do not need to waste CPU copying the entire applied configuration back to
    // the agent on each checkin; instead, they can simply echo back the index or revision number from the expected message upon successful reconciliation. Configurations in large deployments can be 1MB or more.
```

elastic-agent-client.proto
```diff
@@ -284,9 +289,11 @@ message CheckinObserved {
 message ActionRequest {
     // Type of action being performed.
     enum Type {
-        // Custom action (registered by the unit)
+        // Custom action (registered by the unit). Examples include endpoint response actions and OSQuery results.
         CUSTOM = 0;
-        // Defined diagnostics action.
+        // Diagnostics collection action. Implemented by components and units when they which to
+        // include custom information in diagnostics archives. If unimplemented, a ActionResponse
+        // with the FAILED status is required.
```
AndersonQ, line 296: Typo?

```suggestion
        // Diagnostics collection action. Implemented by components and units when they wish to
        // include custom information in diagnostics archives. If unimplemented, an ActionResponse
        // with the FAILED status is required.
```

README.md
rdner: +1 to adding explanation of the `source` field and particularly its purpose, I still don't know why we need to move these large objects without using them.

elastic-agent-client.proto
```diff
@@ -205,21 +201,30 @@ message ComponentLimits {
...
+    // Index or revision of the expected feature flags configuration. When the expected
+    // configuration changes the agent will increment this number and the Features field will be
+    // populated.
     uint64 features_idx = 4;
 
-    // Component is the expected component configuration.
-    // Added on Elastic Agent v8.10.0.
+    // Component is the expected component configuration. Contains configuration expected to apply
+    // globally to the entire component process. Omitted if the client reports it has applied the
+    // current configuration. Added in Elastic Agent v8.10.0.
```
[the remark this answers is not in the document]
rdner: Well, if this statement is not true, let's remove it from the doc.

README.md
cmacknz: It is so new fields can be added without us needing to synchronize them with the release of the agent, they can just be parsed out of the source blob (which is essentially JSON encoded in protobuf).

I'll update the documentation.

elastic-agent-client.proto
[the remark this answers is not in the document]
cmacknz: I was hoping to just fix this, but I likely won't have time so I'll just adjust the docs to reflect what actually happens.

README.md
cmacknz: If we hadn't accidentally released the Beats build with DEV=true previously I'd say yes, but we did do this in the early 8.x releases. Doing that with these would leak credentials in the agent logs unfortunately.

cmacknz: I'll see if I can think of an easy way to turn this on and off.

README.md
faec: This should probably be a conversation elsewhere but this is the first I'm seeing an explicit discussion of the source field so I'll mention it here:

As things stand, it looks like a field at depth n will appear (n+1) times in the config -- once at its actual location, and then one additional time for the source field at each higher level. That's an extremely high overhead in size and complexity... is the payoff worth it? Could we potentially get by with just a single raw copy at the root level instead of recursively reduplicating every nested field? (Is it even possible for the different copies to differ, or is it strictly redundant?)

cmacknz: The goal with the Source field was to ensure that the agent itself doesn't need to be updated if the YAML configuration changes.

At least in the case of the unit configurations this makes sense, since only a sub-set of the fields are required and the rest are unknown to the agent. For feature flags and component configuration the same principle seems to makes sense, but it's less clear if we'd ever use it this way.

The mechanism here was never intended to be efficient, and we've never audited any of this to minimize the size of the configuration.

pierrehilbert:
@cmacknz should we put this one back to draft or is it just something you forgot and we are close to be able to merge?

cmacknz:
I need to come back to this and get it merged, will try this week

cmacknz:
Back to draft while I find time to finish this

cmacknz:
Out of draft. I've added an explanation of the source field, hopefully it makes sense to everyone.

README.md
```diff
@@ -0,0 +1,418 @@
...
+    // CA certificate.
+    bytes ca_cert = 4;
+    // Peer certificate.
+    bytes peer_cert = 5;
+    // Peer private key.
+    bytes peer_key = 6;
+    // Allowed services that spawned process can use. (only used in V2)
+    repeated ConnInfoServices services = 7;
+}
+```
```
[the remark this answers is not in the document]
cmacknz: I don't think I can link to the live contents of a section of a file, I can only permalink to a specific revision. So I'm either pinning to an static version or duplicating the contents here, I duplicated it so it'd be readable in an editor easily.

ycombinator (approved):
Thanks for adding an explanation for the source field.

LGTM!

PR #91, opened 2023-12-27
Add domain socket/named pipe support

aleksmaus:
This adds support for elastic agent client comms over domain socket on *nix and named pipe on windows.
I saw there was a discussion on the agent channel that this would solve some of the issues seen with localhost ports blocked etc. 

There is no changes to the interface signatures, it's still fully backwards compatible. 
Added more tests to cover comms over domain sockets/named pipes with and without TLS.

Can update the agent/beats if this looks ok and work with Endpoint team to get er done.


leehinman:
@aleksmaus I haven't looked deeply at this, but I did notice that you aren't using https://github.com/elastic/elastic-agent-libs/tree/main/api , which is what the rest of elastic-agent uses for socket/named pipe support.  Is there a reason we can't use that?

aleksmaus:
> @aleksmaus I haven't looked deeply at this, but I did notice that you aren't using https://github.com/elastic/elastic-agent-libs/tree/main/api , which is what the rest of elastic-agent uses for socket/named pipe support. Is there a reason we can't use that?

@leehinman I replaced the direct calls to to winio with the github.com/elastic/elastic-agent-libs/api/npipe APIs.

pkg/client/client_v2.go
```diff
@@ -271,13 +273,24 @@ type clientV2 struct {
...
+// npipe:///pipe_name for windows named pipe
 func NewV2(target string, token string, versionInfo VersionInfo, opts ...V2ClientOption) V2 {
 	var options v2options
 	options.maxMessageSize = DefaultMaxMessageSize
 	for _, o := range opts {
 		o(&options)
 	}
 
+	// For compatibility with existing interface the target could contain npipe:// scheme prefix
+	// Set the named pipe dialer option if pipe:// prefix specified on windows
```
blakerouse, line 287:
```suggestion
	// Set the named pipe dialer option if npipe:// prefix specified on windows
```

blakerouse (approved):
Change looks good. Just a typo comment.

PR #93, opened 2024-01-05
Send package.version to components

AndersonQ:
Send package version to components:
 - add AgentInfo to ConnInfo to be sent at component startup
 - rename `ConnInfo` to `StartUpInfo`
 - redefine `CheckinObservedVersionInfo.Version` to be the package version. Before it was the binary build version
 - add BuildHash` to `CheckinObservedVersionInfo` as the hash of the VCS commit the binary was built from.
 - NewV2 has an option to be built with the AgentInfo. 
 - upon receiving AgentInfo, the Version of the component is updated to the package version, which is the agent's version


pkg/client/client_v2.go
```diff
@@ -174,6 +174,9 @@ type V2 interface {
 	//
 	// nil can be returned when the client has never connected.
 	AgentInfo() *AgentInfo
+	// PackageVersion returns the version of the agent package that should be
+	// presented in user visible fields and messages.
+	PackageVersion() string
```
AndersonQ, line 179: I was thinking instead of having this method, the agent could send the whole AgentInfo together with the ConnInfo, that way the client could already populate its local AgentInfo and instead of a new method, the agent info could be used since the very beginning. Without having to wait the 1st check-in.

blakerouse: I prefer that. Have it send the entire AgentInfo in the ConnInfo.

blakerouse (changes requested):

elastic-agent-client.proto
```diff
@@ -323,10 +323,12 @@ message UnitObserved {
 message CheckinObservedVersionInfo {
     // Name of the binary.
     string name = 1;
-    // Version of the binary.
+    // Agent package version sent by the agent.
```
cmacknz, line 326: Thinking a bit more, if this field always needs to equal the agent version there's no value in reporting it. Should we just remove this field entirely?

If we want to keep it, we can have the client code here automatically populate it.

AndersonQ: Yes, it does not provide any value. I'd argue it should be still the binary real version, but this train already left. 
However, keeping the current behaviour does not harm, do not require any change and would allow us in the future to use somehow the real binary version if we ever want. Therefore I'd argue doing nothing is the best option.

elastic-agent-client.proto
```diff
@@ -194,13 +194,16 @@ message UnitExpected {
     UnitLogLevel log_level = 6;
 }
 
-// Only provided on first checkin expected response to the component.
+// Agent information information that the component might want to use for its
```
blakerouse, line 197:
```suggestion
// Agent information that the component might want to use for its
```

elastic-agent-client.proto
blakerouse: Best to keep it for backwards compatibility.

elastic-agent-client.proto
```diff
@@ -455,10 +460,12 @@ enum ConnInfoServices {
...
 
-// Connection information sent to the application on startup so it knows how to connect back to the Elastic Agent.
+// Information sent to component on startup containing the necessary information
+// for the component to connect back to the Elastic Agent and the agent details.
 //
-// This is normally sent through stdin and should never be sent across a network un-encrypted.
-message ConnInfo {
+// This is normally sent through stdin and should never be sent across a network
+// un-encrypted.
+message StartUpInfo {
```
blakerouse, line 468: I like the rename now that its more than just information about the connection.

pkg/client/reader.go
```diff
@@ -64,46 +64,56 @@ func NewFromReader(reader io.Reader, impl StateInterface, actions ...Action) (Cl
...
 	}
-	err = protobuf.Unmarshal(data, connInfo)
+	err = protobuf.Unmarshal(data, info)
 	if err != nil {
 		return nil, nil, err
 	}
-	if connInfo.Services == nil {
+
+	if info.AgentInfo != nil {
+		ver.Version = info.AgentInfo.Version
```
blakerouse, line 78: Why is the version being overwritten? This version information is what is reported to the Elastic Agent. I don't think we want to overwrite this information. This information is used for diagnostics and without it we will not be able to get the real version of the component.

I think you are mixing two things here.

1. The version the component should report in its events
2. The version of the component reported to Elastic Agent through the control protocol.

This change should not effect "2" it should only affect "1".

blakerouse (changes requested):
I think the overriding of the version in the client is incorrect. See my inline comment.

pkg/client/reader.go
cmacknz: > The version of the component reported to Elastic Agent through the control protocol.

The version of the component is now uniquely identified using the VCS hash. 

Let's imagine the latest stack release is 8.12.0. 
With the independent releases project we will have situations where the 8.12.1 build of beats is released to fix a bug in 8.12.0 through an 8.12.0+build20240130 versioned agent. We intentionally want the version everywhere it could appear to be 8.12.0+build20240130. There are two reasons for this:

1. We do not want product level confusion about the appearance of the 8.12.1 version before the official 8.12.1 stack release happens.
2. The 8.12.1 version is going to appear in the 8.12.0+build20240130 and the 8.12.1 release and no longer uniquely ties back to a commit on its own.

This change is just doing the work for the component of ensure that the agent package version 8.12.0+build20240130 overrides the stack version in the component, because when releasing patches outside the stack cadence it no longer has the meaning it did before.

This is because the 8.

cmacknz: For context bump the version in the 8.12 branch to 8.12.1 the day 8.12.0 is released, which is what allows the 8.12.1-SNAPSHOT to be produced. This means any commit released from 8.12 from that point has the 8.12.1 version, unless we override it like this.

AndersonQ: Given the current decision to do not expose the original binary version, as Craig said. It needs to be overwritten here because the component does not know the package version at this point, so it's passing in its real version.

We could on the other hand keep it and when producing the CheckinObserved use the agentInfo version.

Whereas I like having the binary version, is it's not being used, it will only cause confusion be a mistake waiting to happen, at some point someone using it instead of the agentInfo.Version and er end up with a release with mixed versions. Better avoid it at all.

As the field is being kept, if we ever decide to change it, the protocol will still support it.

blakerouse: I am completely lost with that explanation @cmacknz. I am struggling to understand the explanation so that makes me worried to agree or disagree.

If I am understanding it correctly (probably not), we would ship a 8.12.1 filebeat with a 8.12.0 Elastic Agent correct? If that is correct wouldn't we want the diagnostics of a running Elastic Agent with version ` 8.12.0+build20240130` to reporting in the diagnostics that its running a 8.12.1 filebeat? I don't know why we would not want that.

@AndersonQ `VersionInfo` is not a pointer, but is sent by value into the client. So updating that version does not affect the caller, it only affects what is happening inside of the client (so other cannot use it). I still feel we want the actual version reported to the Elastic Agent.

blakerouse: @AndersonQ I spoke with @cmacknz. We agreed that its best to just remove the version string being reported to the Elastic Agent completely.

So please update `VersionInfo` to remove the `Version` field as well as remove it from the protocol.

pkg/client/client_v2.go
```diff
@@ -278,7 +289,12 @@ func NewV2(target string, token string, versionInfo VersionInfo, opts ...V2Clien
 		o(&options)
 	}
 
+	if options.agentInfo != nil {
+		versionInfo.Version = options.agentInfo.Version
+	}
```
blakerouse, line 294: This should be removed. Remove `Version` from `versionInfo`, its no longer used so we should not have any clients set it.

pkg/client/client_v2.go
```diff
@@ -527,9 +543,9 @@ func (c *clientV2) sendObserved(client proto.ElasticAgent_CheckinV2Client) error
 	}
 	if !c.versionInfoSent {
 		msg.VersionInfo = &proto.CheckinObservedVersionInfo{
-			Name:    c.versionInfo.Name,
-			Version: c.versionInfo.Version,
-			Meta:    c.versionInfo.Meta,
+			Name:      c.versionInfo.Name,
+			Meta:      c.versionInfo.Meta,
+			BuildHash: c.versionInfo.BuildHash,
```
blakerouse, line 548: Not used here so it should not be present.

pkg/client/reader.go
blakerouse: I was expecting this to be gone?

blakerouse (changes requested):

PR #95, opened 2024-01-10
Add go vet to CI pipeline and fix reported issues

AndersonQ:


pchila (commented):
Change looks fine, I am worried about the test failure though...

pkg/client/unit_test.go
```diff
@@ -79,8 +79,7 @@ func TestUnitUpdateState(t *testing.T) {
 }
 
 func TestUnitUpdateAPMConfig(t *testing.T) {
-
-	var initialUnitNoAPM = Unit{
+	initialUnitNoAPM := &Unit{
```
pchila, line 82: This struct is now shared among different testcases which makes the test fail when all the testcases are run. Just tried to nest it at the cost of some duplication but that take care of the failures without reintroducing the `go vet` violations

pchila: Added a new commit where we create a new unit per testcase

AndersonQ: I was pushing a fix as well, just copying and pasting it all over. Yours is more elegant. Thanks

PR #96, opened 2024-01-10
Refactor TestClientV2_Checkin_Initial and add test for package version

AndersonQ:
- remove test for wrong token
 - add check to ensure the client haven't returned any error
 - add check for the new BuildHash field of proto.CheckinObserved
 - check AgentInfo is correctly set

### DIFF
--- a/pkg/client/client_v2_test.go
+++ b/pkg/client/client_v2_test.go
@@ -141,69 +141,79 @@ func TestClientV2_Checkin_Initial(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		t.Run(tc.name, func(n string, sc, cc credentials.TransportCredentials) func(t *testing.T) {
-			return func(t *testing.T) {
-				testClientV2CheckinInitial(t, n, sc, cc)
-			}
-		}(tc.localRPC, tc.serverCreds, tc.clientCreds))
+		t.Run(tc.name, func(t *testing.T) {
+			testClientV2CheckinInitial(
+				t, tc.localRPC, tc.serverCreds, tc.clientCreds)
+		})
 	}
 }
 
 func testClientV2CheckinInitial(t *testing.T, localRPC string, serverCreds, clientCreds credentials.TransportCredentials) {
 	var m sync.Mutex
 	token := mock.NewID()
-	gotInvalid := false
-	gotValid := false
 	unitOneID := mock.NewID()
 	unitTwoID := mock.NewID()
 	wantFQDN := true
-	reportedVersion := VersionInfo{}
+
+	packageVersion := "8.13.0+build20060102"
+	wantBuildHash := "build_hash"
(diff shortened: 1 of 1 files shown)


elastic-agent-client.proto
```diff
@@ -323,10 +323,12 @@ message UnitObserved {
 message CheckinObservedVersionInfo {
     // Name of the binary.
     string name = 1;
-    // Version of the binary.
+    // Agent package version sent by the agent.
     string version = 2;
     // Additional metadata about the binary.
     map<string, string> meta = 3;
+    // VCS commit hash of the binary.
+    string build_hash = 4;
```
cmacknz, line 331: Since this has the changes from https://github.com/elastic/elastic-agent-client, shouldn't it be targeted to merge to the AndersonQ:beats-report-package-version branch so we don't see both changes in the PR?

pkg/client/client_v2_test.go
```diff
@@ -91,59 +91,57 @@ func rejectingListener(listener net.Listener) {
...
 		CheckinV2Impl: func(observed *proto.CheckinObserved) *proto.CheckinExpected {
 			m.Lock()
 			defer m.Unlock()
+			if !assert.Equal(t, token, observed.Token,
+				"client sent the wrong connection token") {
+				// returning nil makes the server to disconnect
+				return nil
+			}
 
-			if observed.Token == token {
```
cmacknz, line 105: Why can this be removed?

cmacknz: Nevermind

pkg/client/client_v2_test.go
```diff
@@ -209,51 +187,48 @@ func TestClientV2_Checkin_Initial(t *testing.T) {
...
-	assert.True(t, agentInfo.Snapshot)
+	assert.Empty(t, err, "client returned errors")
 
-	assert.Equal(t, wantFQDN, gotFQDN)
+	require.NotNil(t, agentInfo,
+		"agent info is nil, it should have been set during check-in")
+	assert.Equal(t, agentInfo.Version, packageVersion)
+	assert.True(t, agentInfo.Snapshot, "agent version is a SNAPSHOT")
+
+	assert.True(t, gotFQDN, "FQDN should be enabled")
```
cmacknz, line 221: There should probably be an assert.Len(t, units, 2) right after this since the block above that did this was removed. The test would fail without this, just with an index error below.

```go
if len(units) < 2 {
     return fmt.Errorf("never received 2 units")
}
```

elastic-agent-client.proto
AndersonQ: If I'd have done so, the PR would be on my fork only, not here.

cmacknz (approved):
Found the diff a bit hard to follow, but changing the versions and build hash in the test causes the expected failures so I know its tested what it is supposed to.

PR #97, opened 2024-01-16
Add repeat padding to message chunking

### LINKED ISSUE #94: CheckinV2 message chunking doesn't account for the overhead protobuf adds for repeated structs
Currently `chunk.Expected` and `chunk.Observed` doesn't take into account that each unit has a little more overhead because of how the size is calculated go through each unit. Overhead needs to be added roughly about 4 bytes to ensure that a chunk message still doesn't end up bigger than the maximum message size.

blakerouse:
## Overview

Adds padding to repeating units to ensure that they fit with inside the maximum size. Without the padding the resulting message size can be too large.

# Relates

- Closes #94


pkg/client/chunk/expected_test.go
```diff
@@ -333,6 +335,53 @@ func TestRecvExpected_Timestamp_Restart(t *testing.T) {
...
+				Type: "testing",
+				Name: "testing",
+			},
+			LogLevel: proto.UnitLogLevel_ERROR,
+		}
+		units = append(units, unit)
+		unitsSize += gproto.Size(unit)
+		nextUnitID++
+
+		if unitsSize > minimumMsgSize {
```
AndersonQ, line 363: [Suggestion]
you could have made it part of the for condition

```go
for ; unitsSize > minimumMsgSize ; {
```

pkg/client/chunk/observed_test.go
```diff
@@ -341,6 +343,50 @@ func TestRecvObserved_Timestamp_Restart(t *testing.T) {
...
+			Type:           proto.UnitType_INPUT,
+			State:          proto.State_HEALTHY,
+			Message:        fmt.Sprintf("fake-input-%d is healthy", nextUnitID),
+			ConfigStateIdx: 1,
+		}
+		units = append(units, unit)
+		unitsSize += gproto.Size(unit)
+		nextUnitID++
+
+		if unitsSize > minimumMsgSize {
```
AndersonQ, line 366: same here

```go
for ; unitsSize > minimumMsgSize ; {
```

AndersonQ (approved):

pkg/client/chunk/expected_test.go
blakerouse: fixed

pkg/client/chunk/observed_test.go
blakerouse: fixed

PR #98, opened 2024-01-16
Fix issue with npipe import on Windows.

blakerouse:
Fixes issue in PR that didn't actually import npipe from the elastic-agent-libs.

### DIFF
--- a/pkg/client/util_windows.go
+++ b/pkg/client/util_windows.go
@@ -7,6 +7,8 @@
 
 package client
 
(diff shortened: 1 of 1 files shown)


cmacknz (approved):

aleksmaus:
> Fixes issue in PR that didn't actually import npipe from the elastic-agent-libs.

Thanks!
Sigh, things you miss when you don't have windows dev box.

PR #106, opened 2024-03-26
Add checkin_pid to CheckinObserved

fearful-symmetry:
Continuation of https://github.com/elastic/elastic-agent/issues/4083

In order to get elastic-agent to monitor endpoint, we need to know the pid of endpoint, which we currently can't do, as Endpoint operates a bit more independently than the other beats that are talking to agent. Endpoint can restart independently, and Agent can't track the PID on its own.

This adds a `checkin_pid` integer to `CheckinObserved`. I put it here for two reason:
1) I can't think of anywhere else it could go
2) If endpoint restarts on its own, `CheckinObserved` seems like the most logical place to put a new PID.


elastic-agent-client.proto
```diff
@@ -365,6 +365,10 @@ message CheckinObserved {
     // Supports provides information to the agent about extra features this client supports. Should always be included
     // on first checkin, and not again unless upon reconnect.
     repeated ConnectionSupports supports = 8;
+
+    // Optional field allowing the client to report its own PID. This is useful for applications like endpoint,
+    // where agent has less control over the binary.
+    uint64 checkin_pid = 9;
```
cmacknz, line 371:
```suggestion
    uint64 pid = 9;
```

The checkin is redundant I think, in some languages this will generate a getter like `checkin_observed().checkin_pid()`.

brian-mckinney (commented):
👍

cmacknz (approved):

PR #110, opened 2024-05-22
Add management mode enum to `AgentInfo` metadata 

fearful-symmetry:
Needed for https://github.com/elastic/beats/pull/39403

This adds a `AgentManagedMode` enum to the `AgentInfo` struct. This is in service of having beats send the fleet mode upstream via the user agent. Right now, management mode is only known to elastic-agent, so that data has to travel from Agent to Beats.

This also tinkers with the magefile, as the `go-licenser` install was completely broken for me. This should be correct?


elastic-agent-client.proto
```diff
@@ -107,6 +107,12 @@ enum UnitLogLevel {
     TRACE = 4;
 }
 
+// Reports the mode agent is running in
+enum AgentManagedMode {
+    FLEET = 0;
+    SELFMANAGED = 1;
```
ycombinator, line 113: I don't know if these values are being used somewhere already — hopefully not — but it would be good to name them consistently with existing verbiage for these concepts.
```suggestion
    MANAGED = 0;
    STANDALONE = 1;
```

ycombinator:
> This also tinkers with the magefile, as the go-licenser install was completely broken for me. This should be correct?

Are we sure we need these changes? Why did this break only now?  What did you attempt to do when it broke and what was the symptom — I can try to reproduce this on my end (from `main`).  Also, if we do need these changes, they should go into their own PR as they're unrelated to the topic of this PR.

fearful-symmetry:
@ycombinator Sorry, should have explained that a bit more:

 the `go get` operation we're using breaks on 1.22:
```
mage prepare:all                                                                                                                                                
go: modules disabled by GO111MODULE=off; see 'go help modules'
Error: running "go get github.com/elastic/go-licenser" failed with exit code 1
```

I'm guessing the original code did this since it was a bit of a hack to install binaries without using `go install` However, this doesn't work in go 1.22 anymore:
https://tip.golang.org/doc/go1.22

> go get is no longer supported outside of a module in the legacy GOPATH mode (that is, with GO111MODULE=off). Other build commands, such as go build and go test, will continue to work indefinitely for legacy GOPATH programs.

I figured it was a small enough change that I could just throw it in here, but we can spin it out into a different PR if you want.

PR #125, opened 2024-11-26
fix: use proto.Equal to compare protobuf messages

kruskall:
replace reflect deepequal with proto.Equal

reflection is comparing some internal fields used during marshaling, causing race conditions in the beats reloader code.
the race condition causes APM Server tests to fail with the race detector enabled

### DIFF
--- a/pkg/client/unit.go
+++ b/pkg/client/unit.go
@@ -5,7 +5,6 @@
 package client
 
 import (
-	"reflect"
 	"sync"
 
 	gproto "google.golang.org/protobuf/proto"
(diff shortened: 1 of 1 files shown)


cmacknz (approved):

cmacknz:
I don't think this code is used anymore, but it was wrong before regardless.

cmacknz:
Nevermind, this is in unit.go not the v1 client.

cmacknz:
> reflection is comparing some internal fields used during marshaling, causing race conditions in the beats reloader code.

A test to catch this would look like marshalling and then unmarshalling one of the messages under test before using it in the test to populate all the internal fields to get the comparison to fail. Hard to know to do this until you have been burned by this at least once.

Issue #120, opened 2024-07-22
Run unit tests on Windows.


cmacknz:
Follow up from https://github.com/elastic/elastic-agent-client/pull/119 which is the first change to have OS specific functionality, and discovered that CI doesn't run on a Windows agent.

Make it so that lint+build+test happens on Windows workers as well as Linux.

cmacknz:
CC @rowlandgeoff hopefully something eng prod can take care of.

rowlandgeoff:
> CC @rowlandgeoff hopefully something eng prod can take care of.

Added to the Eng Prod backlog

v1v:
There are test failures when running on Windows:


```
expected_test.go:335: 
        	Error Trace:	D:/a/elastic-agent-client/elastic-agent-client/pkg/client/chunk/expected_test.go:335
        	Error:      	Should be empty, but was   (*proto.CheckinExpected)(Inverse(protocmp.Transform, protocmp.Message{
        	            	  	"@type":         s"proto.CheckinExpected",
        	            	  	"component_idx": uint64(3),
        	            	  	"features_idx":  uint64(2),
        	            	  	"units": []protocmp.Message{
        	            	  		{
        	            	  			"@type":            s"proto.UnitExpected",
        	            	  			"config":           protocmp.Message{"@type": s"proto.UnitExpectedConfig", "id": string("testing"), "name": string("testing"), "type": string("testing")},
        	            	  			"config_state_idx": uint64(1),
        	            	- 			"id":               string("second-one"),
        	            	+ 			"id":               string("first-one"),
        	            	  			"state":            s"HEALTHY",
        	            	  			"type":             s"OUTPUT",
        	            	  		},
        	            	+ 		s`{id:"first-two"  state:HEALTHY  config_state_idx:1}`,
        	            	  		{"@type": s"proto.UnitExpected", "config_state_idx": uint64(1), "id": string("second-two"), "state": s"HEALTHY"},
        	            	  	},
        	            	  }))
        	Test:       	TestRecvExpected_Timestamp_Restart
--- FAIL: TestRecvExpected_Timestamp_Restart (0.00s)
=== RUN   TestRecvExpected_RepeatPadding
--- PASS: TestRecvExpected_RepeatPadding (3.69s)
=== RUN   TestObserved
=== RUN   TestObserved/unit_too_large_to_fit
=== RUN   TestObserved/first_chunk_too_large
=== RUN   TestObserved/chunk_checkin_message
[...]

PR #130, opened 2025-03-27
github-actions: for the CI

v1v:
### What

Support:
- lint on linux and windows
- test on linux and windows
- update

Add dependabot

Use github composite action for setting up the environment

Skip failures on `Windows-2025`

### Why

This project uses an old version of golang/magefile that requires further changes or an update in the BK images. See https://github.com/elastic/elastic-agent-client/pull/129

It seems more trivial and easy to use GH actions instead. See [`8784ed3` (#130)](https://github.com/elastic/elastic-agent-client/pull/130/commits/8784ed35270d91685ba28353106468a18871ff0a) versus https://github.com/elastic/elastic-agent-client/pull/130/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f

Faster builds:

image

vs

image

and it added the OS windows

Part of https://github.com/elastic/elastic-agent-client/issues/120


oakrizan (approved):

pierrehilbert (approved):

.github/actions/setup/action.yml
```diff
@@ -0,0 +1,21 @@
...
+        if: runner.os != 'Windows'
+        with:
+          go-version-file: go.mod
+          cache: true
+      # See https://github.com/actions/setup-go/pull/515
+      - uses: antontroshin/setup-go@bda02de8887c9946189f81e7e59512914aeb9ea4
+        if: runner.os == 'Windows'
+        with:
+          go-version-file: go.mod
+          cache: true
```
swiatekm, line 18: From what I understand, the only thing this fork does is change the location of GOCACHE and GOMODCACHE. Wouldn't it be simpler to use the original action, and just set these variables ourselves?

swiatekm (approved):

.github/actions/setup/action.yml
v1v: I applied the same pattern we did also for the apm-server at:
- https://github.com/elastic/apm-server/pull/15623

No strong opinion here.

Maybe that could be done in a follow-up?

